PR #294, opened 2014-10-04
add sticky table header on grid page

idealatom:
I think sticky table header is more convenient, especially on big scrollable grids


pydanny:
This is AWESOME!

I apologize for how long it's taken to merge this in.

PR #330, opened 2015-10-05
Fix #321

### LINKED ISSUE #321: dupe in grid
https://www.djangopackages.com/grids/g/countries/

django-airports 	django-airports 

Not sure if this is a data problem or a code problem.

radiac:
This fixes #321 

The ``_attach_grid.html`` template was included twice by package.html when the package was on no grids. The js in ``ajax_grid_list.html`` then submitted both within a few ms, triggering a race condition in ``grid.views.add_grid_package`` which meant it was sometimes added twice.

This change removes the grid search form from the panel title and always includes the panel body form (permission permitting). The search form now has a title to improve layout, and the help text has been corrected.

### DIFF
--- a/templates/grid/ajax_grid_list.html
+++ b/templates/grid/ajax_grid_list.html
@@ -2,7 +2,7 @@
 <div id="target">
     <h2>{% if grids.count %}{% trans "Grids" %}{% else %}{% trans "No grids found" %}{% endif %}</h2>
     {% if grids.count %}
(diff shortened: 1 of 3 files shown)


radiac:
Build is failing because base branch is failing - errors are the same.

pydanny:
Great detective work! I'll merge it in and fix the problem with the test myself. :smile:

PR #340, opened 2016-06-03
Fixed copyright year in footer

bmispelon:
I noticed that the year in the footer was out of date so I updated it.

I fixed the templates when I could so that it now automatically displays the current year.

However, I was not able to do that in the static HTML pages (error.html and maintenance.html) so on those pages I just incremented the year.

I also tried to use a dynamic year in the sphinx configuration file but I'm not sure if that's a valid solution.

Thanks!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,7 +11,7 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys, os
+import sys, time, os
 
 # todo: see if VIRTUAL_ENV variable is in the paths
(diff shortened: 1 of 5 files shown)


pydanny:
Could you merge in the latest and I'll accept this pull request?

bmispelon:
@pydanny done :sparkles:

pydanny:
Hmmm... tests are failing. I'll take a look at it today.

pydanny:
The test failure is unaffiliated with this pull request so I'm brought this in anyway.

PR #451, opened 2017-03-24
XSS flaw bugfix

asteinhauser:
There is no server-side check that the pypi_url does not contain an arbitrary malicious JavaScript. Therefore the attacker can plant a stored XSS flaw as demonstrated in here:

https://djangopackages.org/packages/p/test/

### DIFF
--- a/templates/package/package.html
+++ b/templates/package/package.html
@@ -77,7 +77,7 @@
      <div class="panel-heading clearfix">
        <h3 class="panel-title pull-left">PyPI</h3>
      </div>
-     {% if package.pypi_url and package.pypi_url != "http://pypi.python.org/pypi/" %}
+     {% if package.pypi_url and package.pypi_url != "http://pypi.python.org/pypi/" and package.pypi_url|slice:":4" == "http" %}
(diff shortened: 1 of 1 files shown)


jayfk:
```{{ package.pypi_url }}``` should be escaped automatically by Django, except called explicitly with `|safe`: https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#safe

asteinhauser:
Sure. It is a context-sensitive XSS flaw. It is escaped by HTML sanitizer (the Django autosanitizer), but the autosanitizer does not handle dangerous URLs. It fixes only <, >, ", ' and & values. So the attacker can put there javascript: URL and append any JavaScript code which will be executed when the user clicks on the link - as I demonstrated on the testing package. For a proper sanitization on place, you would need a filter like urlencode, but that would break the application logic.

jayfk:
Yep, you are right. Just saw this mail flying by and I wasn't looking at the content.

I'm going to merge this in. For a number of sad reasons, the repo is currently in a split brain mode. I have to apply this manually on the old server.

PR #467, opened 2017-04-17
Port to Django 1.11

jayfk:
Runs, but tests are failing miserably.

The biggest problem is that some passwords are using (now unsupported) raw password hashes in fixtures.


jayfk:
Down to 3 failing tests 👏

jayfk:
Merged in https://github.com/djangopackages/djangopackages/pull/498

PR #482, opened 2017-08-25
Fix URL for badges

amitness:


amitness:
@webyneter Please check the PR and merge it.

PR #495, opened 2017-11-30
Add item pk to Search model

kandziu:
I think that used DRF for new API version is greats idea, but new controllers needs pk and in old versions used slug. I'm upgraded **searchv2** and include **item_pk** to APIv4 search response.


jayfk:
Thanks! 

I'll probably need some time to review this one, hang in there :).

kandziu:
hi @jayfk, i fixed all merge conflicts

jayfk:
Could you tell me more about the reasons for this PR? Maybe I'm missing something obvious.

Why do we need that? What problem does it solve?

kandziu:
Search result not contains items pk but this need for api v4

pydanny:
Still haven't explained what problem it solves.

kandziu:
Detail views in the api v4 accepts _pk_ like `..api/v4/packages/200/`, but search response `..api/v4/search/?q=django` not contains _pk_ and this solution add _pk_ into search response items.
Fixed management `NoArgsCommand` class, clear _pdb_ breakpoints and `get_resource_uri` method as well.

kandziu:
Hello @jayfk, @pydanny. What do you think about this PR? If **djangopackages** does not need this i could close PR.

PR #512, opened 2018-04-30
Fix load_dev_data command

joshua-s:
In the current version, this command produces several errors due to outdated imports and other issues.

Not quite ready yet.


vaibhav-jain:
@joshua-s I think this has been addressed in #560 closing this for now re-open it if you think I missed something

joshua-s:
Thank you!

Issue #517, opened 2018-08-17
Vulnerability: The XSS without HTML: Client-Side Template Injection with AngularJS


noisy:
<REDACTED>

pydanny:
Please don't post the details of security vulnerabilities publicly.

vaibhav-jain:
closing as per https://github.com/djangopackages/djangopackages/pull/518#issuecomment-643776006

PR #518, opened 2018-08-17
Fix for #517, XSS Vulnerability

noisy:
Please check that carefully. I am not 100% sure, that this is a proper fix.


vaibhav-jain:
Hi @noisy

thanks for the fix but tested and I think the site is safe, if you still feel that we need a fix, please reach out to me and we can test it again

Issue #448, opened 2017-03-15
Hide abandoned packages from the Grid view


adamn:
I've been using Django for many years and was recently showing it to a new developer.  When showing how great it is I realized that there is a bunch of cruft on the site now.  Some packages haven't been updated in years and should not be shown to users looking for a solution to their problem.

I wonder if it makes sense to have an `in_grid` boolean for the `Package` model which defines whether the package shows up in the grid.

An inactive package should still have it's individual package url but I think the grids would be cleaner and the site itself would be more delightful.

An easy heuristic to do the initial cull is via Python3 support with maybe 1 or 2 manual exceptions.  Any package that doesn't yet have Python3 support is probably dead and likely won't work with Django 1.10, let alone 1.11 (and certainly not Django 2 of course).

This also means that the Python 3 support row can be removed from the grids as well - further cleaning things up.

fabiocaccamo:
Any news about this request?

There is no reason to show packages which are not updated since more than 2 years and doesn't provide python 3 support. They could be hidden, or moved at the end of the list.

The last update date should have more weight in the ordering, I prefer to try a new package with 10 stars vs an abandoned one with 1k stars.

RamezIssac:
I think an approach for this can the ordering of the packages ..
The packages are ordered by the github stars (as far as i see).

Maybe it should ordered be a function of the stars and the last update date

Issue #507, opened 2018-04-09
Python 3 support seems to depend on PyPI presence


patrys:
I am the maintainer of [Saleor](https://djangopackages.org/packages/p/saleor/). Saleor is Python 3 only. There's no way for me to indicate that on Django Packages. The project is not pip-installable so it does not have a `setup.py`.

NyanKiyoshi:
Bump.

vaibhav-jain:
@patrys just checking is it still an issue?

patrys:
Yes, it's still an issue. If you look at the [e-commerce grid](https://djangopackages.org/grids/g/ecommerce/) you will see that Saleor is listed as not supporting Python 3 despite the fact that the project itself only supports Python 3.7+.

NyanKiyoshi:
We are still unable to specify classifier information without PyPi presence

jefftriplett:
Closing this issue because the metadata now exists on PyPI. The better fix is a pretty massive cleanup of our existing data and then changing some assumptions around that Packages should be assumed Python 3 by default. Then we can change how scoring works with Python 2 only packages.

PR #537, opened 2019-05-10
Grid with python3 supported packages only

rationalbeing:
Resolved Issue #448 - Filtered Python 3 supported packages in grid

### DIFF
--- a/grid/views.py
+++ b/grid/views.py
@@ -332,6 +332,9 @@ def grid_detail(request, slug, template_name="grid/grid_detail.html"):
     features = grid.feature_set.select_related(None)
 
     grid_packages = grid.grid_packages.order_by("-package__repo_watchers")
+    python3_supported_grid_packages = [
+        gp for gp in grid_packages if gp.package.last_released() and gp.package.last_released().supports_python3
+    ]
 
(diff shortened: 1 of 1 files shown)


vaibhav-jain:
@rationalbeing Could you add testing steps as well?  they will be highly useful for historical purposes

vaibhav-jain:
@rationalbeing Could you add screenshots of the UI before and after this PR how it is affecting?

jefftriplett:
One thing to note is that we don't have a universally reliable way of telling that an App is using Python3 if it's not in PyPi. Check out #507 for an overview of this. We need to roll out a model change to allow us to overwrite it.

jefftriplett:
We now have updated classifier support in #683 which gets us one more PR closer to having slightly more accurate data to judge this by. For categories of projects which might not be in pypi, I think we still need a fix to avoid accidental removal.

jefftriplett:
This PR is depreciated by #700 and #702 

@rationalbeing thank you for your contribution. I'm closing this out because we need to do a little more work both in #702 and accounting for framework projects that aren't in pypi that are still py3 compatible. Instead of filtering out py3 projects, we are now penalizing more heavily than we were before #700

PR #558, opened 2019-10-19
Update django to 2.2.5

rexhepberlajolli:
Since django 1.11 LTS is going to end in a couple of months, project needs the new LTS version to work properly. This PR has changes required to keep the project running after updating to django 2.2


vaibhav-jain:
HI @rexhepberlajolli,

Changes look good to me but I'll take a pull and test them locally.. Upgrading the `dj-pagination` will also solve the issue we are facing while `RTD` build as it is trying to install `django>1.8` but we are not currently supporting that.

 Good work 😄

rexhepberlajolli:
@vaibhav-jain I fixed the requests and debug toolbar warning, also updated the floppy-forms to 1.8.0 (finally we got a release). The warning coming from floppy-forms is not on our end currently and there is an open issue about it https://github.com/jazzband/django-floppyforms/issues/189.

I can filter the warning on `manage.py` or we can leave it until there is a newer release that fixes it, let me know your thoughts on this.

vaibhav-jain:
@rexhepberlajolli looks great. **LGTM**

PR #559, opened 2019-10-28
Remove whitenoise from requirements since we are not using it

rexhepberlajolli:
White noise causes pyup to fail due to a security update. Since we are not using it this PR removes it


vaibhav-jain:
Yep @rexhepberlajolli it should be removed. **LGTM**

PR #568, opened 2020-01-15
Added target blank to repo and pypi links

Frohus:
Added target="_blank" to PyPi and Repo links in the package view. 
Currently browsing packages is a bit annoying when you browsing more of them and you constantly need to right click and open in the new tab.


vaibhav-jain:
It is a nice feature to have tested and it is working as expected. LGTM

PR #576, opened 2020-03-26
docs: Fix simple typo, seperated -> separated

timgates42:
There is a small typo in docs/repo_handlers.rst, package/repos/base_handler.py.

Should read `separated` rather than `seperated`.


vaibhav-jain:
LGTM thanks @timgates42 for the submission.

PR #619, opened 2021-07-27
Upgrade PSA

snmishra:
The old version doesn't work with GitHub in dev (may be version in production is grandfathered?)


jefftriplett:
@snmishra thank you for your contribution. This was inadvertently upgraded in #621 when I switched over from pip-tools, and I will work on deploying it this weekend.

snmishra:
@jefftriplett That's even better.

PR #634, opened 2021-08-03
Remove erroneous debug=false

davidfischer:


jefftriplett (approved):
🤦  to me but thank for spotting it quickly and the fix.

PR #678, opened 2021-08-08
Reorganizes README badges

sobolevn:
Changes:
1. Removes `Python3 blockers` badge, like this project is already using Python3, why does it show any blockers?
2. Puts more important badges first, then PyUP with less important info


jefftriplett:
@sobolevn thank you for the contribution! Much appreciated!

Issue #676, opened 2021-08-08
⬆️ Update Trove Classifier for Licenses


jefftriplett:
A good first issue might be upgrading our list of licenses from a hard-coded list in `settings/base.py` (see LICENSES or `settings.LICENSES` references). This will help settle any licensing drift/misclassification if it exists, keep us up-to-date with low effort, and give us a library to work with future features and upgrades. 

The PyPA built the [trove-classifiers](https://github.com/pypa/trove-classifiers/) library to help us out. 

## Steps to complete

1. Add trove-classifiers to our `requirements.in` and `requirements.txt` (`pip install pip-tools` and see the `pip-compile` command at the top of requirements.txt for the format to rebuild the file).
2. Rebuild our container (`docker-compose -f dev.yml build`)
3. Update `package/utils.py` to import our new library. Add `from trove_classifiers import deprecated_classifiers` towards the top of the file.
4. Update the [`normalize_license`](https://github.com/djangopackages/djangopackages/blob/ea080a17560bfefa73fb43bc577f63b4aeb2299b/package/utils.py#L69) to reference `classifiers` instead of `settings.LICENSES`.
5. Remove `LICENSES` from `settings/base.py` https://github.com/djangopackages/djangopackages/blob/ea080a17560bfefa73fb43bc577f63b4aeb2299b/settings/base.py#L336-L402
6. Run tests and submit a PR if everything passes. 

----

## Bonus

Update the `normalize_license()` function to limit the `classifiers` to only look for items that start with `License` to prevent leaking non-licenses types in our comparison.

jefftriplett:
@ShreehariVaasishta let me know if this looks like something you want to take on? Let me know if the pip-tools or any other step doesn't make sense.

ShreehariVaasishta:
Yes, will work on it.

jefftriplett:
@ShreehariVaasishta good work. Thanks again for this 👍

jefftriplett:
@ShreehariVaasishta re-opened for the bonus section. Sorry about that.

ShreehariVaasishta:
thank you for re-opening this issue.

jefftriplett:
fixed in #734

PR #683, opened 2021-08-14
Update trove classifiers

ShreehariVaasishta:
This pull request refers to the issue https://github.com/djangopackages/djangopackages/issues/676.


package/utils.py
```diff
@@ -1,3 +1,4 @@
+from trove_classifiers import deprecated_classifiers
```
jefftriplett, line 1: I think we want: 

```suggestion
from trove_classifiers import classifiers
```

package/utils.py
```diff
@@ -66,7 +67,7 @@ def normalize_license(license):
     """
     if license is None:
         return "UNKNOWN"
-    if license.strip() in settings.LICENSES:
+    if license.strip() in deprecated_classifiers:
```
jefftriplett, line 70: See above change too: 

```suggestion
    if license.strip() in classifiers:
```

jefftriplett (changes requested):
This is great. I think we just need to apply the suggestions and see how tests do.

package/utils.py
ShreehariVaasishta: Got it, thanks for the help.

jefftriplett (approved):
🎉 great work. Thank you for your contribution!

PR #699, opened 2021-08-30
698 debug follow up

jefftriplett:
This is an extension of #698 

Fixes  #685


jefftriplett:
Thank you @ShreehariVaasishta, this takes your changes plus a test fix. I couldn't push back to your branch or I would have worked on it there. (GitHub forks ftw)

PR #700, opened 2021-08-30
595 better ranking follow up

jefftriplett:
🎉  thank you  @RamezIssac  for all the help in landing this. 

This is a follow up to #595 to fix: 

- [x] `ModuleNotFoundError: No module named 'dateutil'`
- [x] Migration 0006 was throwing an AttributeError on `calculate_score` not being accessible:
- [x] ~`Package.objects.filter(pk=package.pk).update(score=package.calculate_score())`~ added a calculate_score management command

----

There is more to do here, but this gets enough of the groundwork laid, that I am going to merge and work on some follow up as we start sampling results.


jefftriplett:
🎉  thank you  @RamezIssac  for all the help in landing this.

PR #701, opened 2021-09-06
deprecation

gabbyprecious:


templates/grid/grid_detail.html
```diff
@@ -116,6 +116,7 @@ <h3>{% trans "Features currently being evaluated" %}</h3>
                     {% for grid_package in grid_packages %}
                         <td>
                             {% with grid_package.package.repo_description|truncatewords:20 as short %}
+                                {%if %}
```
jefftriplett, line 119: This looks like it will cause an Exception on the page.

package/models.py
```diff
@@ -67,7 +68,16 @@ class Package(BaseModel):
     documentation_url = models.URLField(_("Documentation URL"), blank=True, null=True, default="")
 
     commit_list = models.TextField(_("Commit List"), blank=True)
-
+    date_deprecated = models.DateTimeField(blank=True, null=True)
+    deprecated_by = models.ForeignKey(User, blank=True, null=True, related_name="deprecator", on_delete=models.PROTECT)
```
jefftriplett, line 72: This is a great addition. Are you integrating it anywhere?

jefftriplett: Follow-up: At a minimum, we should add this field to our raw_id_fields list to avoid the admin getting overwhelmed by 15k users.

jefftriplett (changes requested):
Good start. I added a few comments about a concern or two.

jefftriplett:
Thank you, @gabbyprecious

PR #717, opened 2021-09-24
Automate contributors updates

adellylima:


jefftriplett (approved):
thank you @adellylima, this looks good.

Issue #646, opened 2021-08-06
📦 Package lists can be too long and outdated


jefftriplett:
Twitter feedback via https://twitter.com/JackDLinke/status/1423748550962434055 and https://twitter.com/JackDLinke/status/1423750797247094790

> Is there a plan for filtering/user options? Several of the bigger grids have a bunch of packages that are python2-only, which I would prefer never to see. Same with packages that haven't seen updates in a decade. Hiding those by default would reduce load significantly.

----

> Maybe the forms grid? I think it's one of the largest on the front page.

> https://djangopackages.org/grids/g/forms/

> ~92 packages listed, and MANY of those are VERY out of date or do not seem to have python 3 support.

----

This feels related to https://github.com/djangopackages/djangopackages/pull/537

gabrielolnunes:
Hi @jefftriplett, if you agree with that I can give a try implementing a filter section with a checkbox for **Python 3**.

jefftriplett:
Sounds great. 👍

jefftriplett:
Partially fixed in #719. 

Overall, we need a bit more work with the new `package.Package.pypi_*` fields to help score the packages to see if they are relevant. 

We should penalize packages that have been depreciated from another package.

jefftriplett:
## Audit (March 2025)

Significant progress has been made since this was filed:

### What's been fixed
- **10-package default cap** on grid detail pages with a "View All" link
- **Deprecated/archived packages are excluded** from grids via `Package.objects.active()`
- **Score-based filtering** removes packages below `PACKAGE_SCORE_MIN`
- **Search, sort, and "Stable" filter** available on grid pages
- Forms grid is down slightly from ~92 to 89 packages

### Remaining gaps
- **No "recently active" filter** to hide packages with no updates in N years
- **No visual indicator in grids** for stale-but-not-deprecated packages — the deprecation banner only shows on individual package detail pages
- Large grids still exist (Admin Interface: 127, Forms: 89, NoSQL: 65, Blog: 57) though the 10-package cap mitigates the UX issue

jefftriplett:
I think this is taken care of with our recept redesign.

PR #719, opened 2021-09-25
Issue/646 - Add filters to Grid

gabrielolnunes:
Adds two filters to grid: **python3** and **stable**.
Refers to: #646

**Changelog:**
- Add filter section in the grid template.
- Add JS to handle query strings, filter change and redirect.
- Add filter variant to the cache template tag.
- Add CSS to render the filter section.
- Add filter to grid detail view.
- Add filter whitelist to prevent cache overflow.
- Serialize sorted filters to prevent cache duplication.

**Important Note:**

With those 2 filters, the total of different templates rendered for each grid increases up to four, one for each possible filter combination.

Using the Authentication grid as an example:
1. `python3=False` and `stable=False` - *all = 117 (current)*
2. `python3=True` and `stable=False` - *python3 = 67 (+57%)*
2. `python3=True` and `stable=True` - *stable python3 = 41 (+35%)*
2. `python3=False` and `stable=True` - *stable = 44 (+38%)*

As a result, the **grid template cache might increase up to 130%**.

### DIFF
--- a/grid/views.py
+++ b/grid/views.py
@@ -1,5 +1,7 @@
 """views for the :mod:`grid` app"""
 
+import json
+
 from django.conf import settings
 from django.contrib.auth.decorators import login_required, permission_required
 from django.contrib import messages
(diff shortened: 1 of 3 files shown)


jefftriplett:
@gabrielolnunes this is great. I might need some time to verify that we don't have any caching issues in production before I can merge it. I think that might be a lingering issue.

jefftriplett:
@gabrielolnunes I'd like to update the UI in the future, but this takes the edge off. Thank you again.

Issue #728, opened 2021-10-01
Fix 409 Git Repository is empty.


jefftriplett:
Occasionally a package gets moved or deleted we get a 409 error from github in our logs when the`manage.py package_updater` command is run. 

```
For dwangoapp, <class 'github3.models.GitHubError'>: 409 Git Repository is empty.
```

We should archive these and maybe log a reason so that we can stop checking GH every day for updates. They should also be removed from Grids and possible Search too if we can't locate where they went.

kishan3:
Hi, @jefftriplett I would like to understand which part of the project I should check for this.

jefftriplett:
@kishan3 thanks for the following up. We can use Python logging (should already be imported) and use `logger.error("some meaningful message here")` to log a reasonable error message. 

We can probably add an `active  = BooleanField(default to true)` and toggle it to False when we run into this 409.  

I'll think about this some more and I might do a quick follow up PR because that active bit is going to touch a lot of places.

kishan3:
Thanks, @jefftriplett I created a basic PR You can advise if I need to do something more for this.

jefftriplett:
Fixed in #733

PR #733, opened 2021-10-03
Log a meaningful error message instead of raising an exception.

kishan3:
This should help in addressing #728

### DIFF
--- a/package/management/commands/package_updater.py
+++ b/package/management/commands/package_updater.py
@@ -47,6 +47,7 @@ class Command(BaseCommand):
                     package.fetch_commits()
                 except Exception as e:
                     logger.error(f"Error while fetching package details for {package.title}.")
+                    raise PackageUpdaterException(e, package.title)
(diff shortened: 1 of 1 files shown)


package/management/commands/package_updater.py
```diff
@@ -46,7 +46,7 @@ def handle(self, *args, **options):
                     package.fetch_metadata(fetch_pypi=False)
                     package.fetch_commits()
                 except Exception as e:
-                    raise PackageUpdaterException(e, package.title)
+                    logger.error(f"Error while fetching package details for {package.title}.")
```
jefftriplett, line 49:
```suggestion
                    logger.error(f"Error while fetching package details for {package.title}.")
                    raise PackageUpdaterException(e, package.title)
```

jefftriplett:
@kishan3 thanks for the PR. I made one small change so we could keep the custom exception. thank you again (I still plan to update the first ticket, but just juggling family today)

PR #753, opened 2021-10-29
:fire: Removes dj-pagination

### LINKED ISSUE #749: Modernise tables / dj-pagination is archived
I was looking at some of the other dependencies of the project and noticed that dj-pagination had been [archived](https://github.com/pydanny/dj-pagination).

I'm assuming the project would therefore like to try and remove reliance on this package?

The package is used in a couple of [templates](https://github.com/djangopackages/djangopackages/blob/506184070e37bdcfad5658d8a1579f03db2593aa/templates/package/category.html#L54). I have had a look at what it would take to move to Django's own `Paginator` and this looks simple enough. However it is used in conjunction with `webstack-django-sorting` (wds) which enables you to click on the table headers to sort the data.

Unfortunately the package [doesn't work](https://github.com/webstack/webstack-django-sorting/issues/12) with Django's Paginator. So the complexity here is much higher than "just" a Paginator.

So I guess there are few options:

- Leave things alone until there is an incompatibility with Django (I've had a PR merged to wds to add Django 4.0 support)

- Use Django Paginator, drop the sort feature

- use Django Paginator, roll our own sort feature

- Use another package, `django-tables2` seems to do what we need, it looks like it covers the features of both of the above packages and seems to be well maintained (release on PyPI already declares django 4.0 and py 3.10 support)

https://github.com/jieter/django-tables2

What are your thoughts, there are certainly things I am missing, but to me it feels like Django-tables2 is the front runner here. Not sure what a patch to include it would actually look like though but feels like it is worth investigating. 

On mobile apologies for poor spelling/ grammar etc

jefftriplett:
Fixes #749

- [ ] fix pagination in `category.html` 
- [ ] fix pagination in `python3_list.html`


jefftriplett:
Closing. This is so outdated, it's better to take a fresh approach to it.

PR #755, opened 2021-10-29
Avoid docker image/volume collision by using a namespace

luzfcb:
## Description

- Add namespace to dev docker images and volumes
- Change the Django runserver port to 18000

## Rationale

The basic idea of this pull-request is to avoid having common names for images and volumes on `dev.yml` so that a potential contributor doesn't have to deal with image name collision issues and docker volumes.

It also changed the port used by runserver from 8000 to 18000 for the same reason.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,7 +58,7 @@ To download, install and start the local server for development, simply run::
     docker-compose -f dev.yml build
     docker-compose -f dev.yml up
 
-Then point your browser to http://localhost:8000 and start hacking!
+Then point your browser to http://localhost:18000 and start hacking!
 
 To run tests, run::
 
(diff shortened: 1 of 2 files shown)


jefftriplett:
Sorry for the slow response. I'm semi-back now. One variation we might do would be to: 
- add a `docker-compose.override.yml-dist` file
- add `docker-compose.override.yml` to `.gitignore`
- encourage customizations to: `cp docker-compose.override.yml-dist docker-compose.override.yml`

The addition of `docker-compose.override.yml` should give us room to change ports to be whatever someone wants. 

I'd kind of like to rename `docker-compose.yml` to `docker-compose.prod.yml` or something too so we can make `docker-compose` kind of just work.

jefftriplett:
I'm going to take a merge and update later approach here. We need to update docs either way, but just trying to keep things moving forward.

Issue #720, opened 2021-09-25
✨  Bring back Grid of the Week


jefftriplett:
I was browsing logs and noticed that our top-10 most-searched-for grids weren't marked to be included in the header, so I added them. Now I'm wondering if bringing back the Grid of the Week (gotw) feature wouldn't be a nice way to highlight some of these. 

The homepage exposes a `gotw` variables and at some point, support was pulled for this. Based on git history in the `_homepage.html` template, it looks like we could easily support it again. This is an approximation of what the old theme looked like: 

```html
  <div class="row">
    <div class="col-sm-4 col-lg-4">
      {% if gotw %}
        <!-- start GOTW panel -->
        <h2>{% trans "Featured Grid" %}</h2>
        <h4><a href="{{ gotw.get_absolute_url }}">{{ gotw.title }}</a></h4>
        <p>{{ gotw.description }}</p>
        <p>
          <strong>{% trans "Packages" %}:</strong> {{ gotw.gridpackage_set.count }},
          <strong>{% trans "Features" %}:</strong> {{ gotw.feature_set.count }}
        </p>
        <!-- end GOTW panel -->
      {% endif %}
    </div>
  </div>
```

jefftriplett:
This needs another look but mostly needs a schedule and a plan for announcing weekly grids.

ryancheley:
In order to do this we need to determine if #112 is something that needs to be done OR if we should just keep it as is for now. I'm not sure that the refactor that we'd need to do for #112 would be worth it

ryancheley:
With the closing of #112 I've started working on this. PR coming soon(ish)

ryancheley:
@jefftriplett is there anything left to do for this? I think that 99fda7dc23d completed it

jefftriplett:
It needs some elbow grease and some grid cleanup. Let me push out the last bit of the scoring refactor, and we can try to prioritize this next. 

I also ran the whole website through a bootstrap to tailwind scripts, and the results are good, but not quite good enough to move it over. I think that would help to get over my "why is this so hard to style?" thoughts that are tripping me up.

ryancheley:
@jefftriplett is this something we can sprint on while at DjangoCon US?

jefftriplett:
Closing - Grid of the Week was brought back in #1544.

PR #760, opened 2021-10-29
Grid of the Week panel on the Home page

adellylima:
**Description:**
This PR adds back the Grid of the week panel to the home page.

**_issue:_** https://github.com/djangopackages/djangopackages/issues/720


jefftriplett:
@adellylima thank you for the contribution and for filling in these details. (apologies for the delay. I have been on paternity leave.)

PR #761, opened 2021-10-29
Handle archived projects

cacoze:
**How to test:**
- Run the package updater command, and check on admin if archived GitHub Repos are flagged as archived.
- Access the package detail page and check if you can see a warning informing if that package is archived.

**Description of my solution:**
Fixes #761 
- Added `repo_archived` field to Package model, which will be populated when calling the fetch_metadata method from GithubHandler.(There's no direct field that informs if a GH repository is archived on `github3.py < 1.0 `(however, there is one in the latest version)
- Added a warning bootstrap alert on the package detail page to inform if a package was archived.

**Screenshots:**
image

image


jefftriplett:
This is wonderful and the template visually looks good. 

I would prefer `repo_archived` to be a DateTime field so that we can track when a package was archived. If you have time to swap this then 👍 ; otherwise, I can fork from your PR and swap it around.

cacoze:
I can do that change @jefftriplett, but we don't have that information from GitHub API.  So, I could manually generate a timestamp and fill the DateTime field during the `fetch_metadata` call. Does that sound ok to you?

jefftriplett:
@cacoze you making the change saves me a ton of time. You can default the date to `timezone.now()`. I think we only care that the library has been archived more than than `when`. 

I like having a date on our side so we know when we marked it. Then we can order by this date and use it to look for trends later if we want.

labcodes-selecao:
@jefftriplett  I"ve just pushed the change for `DateTimeField`
Sorry for the delay :v:

cacoze:
@jefftriplett I"ve just pushed the change for `DateTimeField`. Let me know what do you think :v:

jefftriplett (approved):

jefftriplett:
@cacoze thank you for this contribution and your patience. I'm newly off paternity leave and slowly catching up.

PR #762, opened 2021-10-29
Avoid running the docker image building on forks

luzfcb:
Since the deploy workflow creates a commit on the main branch, the current configuration ends up causing the main branch of a fork to diverge from our main branch, which can hinder the contribution flow.

This pull-request makes the workflow deploy only run automatically in our organization, that is, it will be ignored in forks.


jefftriplett (approved):

jefftriplett:
good catch and fix!

PR #770, opened 2022-01-06
✨  Adds Celery support

### LINKED ISSUE #675: 🚜  Migrate to a task queue (Add Celery)
I noticed that the package_updater was running three copies and seemed to be stuck. I noticed that there was a 5-second sleep between every package that is in our database. 

This works as a way to workaround rate limit issues, but we can move to a task queue and avoid blocking via sleep() completely. We should get either 5k (newer APIs might even go as large as 15k an hour) and be able to manage this both from the front and backend. 

I'd like to see the /post-data/ endpoint move the fetching logic to a queued task to cut down on IO too.

jefftriplett:
Fixes #675

This is a fairly MVP setup for Celery to get it deployed then we can migrate either some cronjobs or offload some other blocking tasks to Celery. 

The compose side is managed with profiles which makes running celery optional to our dev setup. 

```shell
# to build
docker-compose --file docker-compose.dev.yml --profile=celery build

# to run
docker-compose --file docker-compose.dev.yml --profile=celery up

# to shutdown
docker-compose --file docker-compose.dev.yml --profile=celery down
```


luzfcb:
@jefftriplett 
@nicysneiros started efforts on this: https://github.com/labcodes/djangopackages/pull/1.

jefftriplett:
@luzfcb thank you for pointing this out. I wasn't seeing the PR previously. I'll pull it in over the next day or so.

jefftriplett:
I'm going to merge and start a new integration branch to pull in https://github.com/labcodes/djangopackages/pull/1 changes

Issue #111, opened 2011-10-02
Limit the size of data going into textfields to 500 characters


pydanny:
If someone needs more than 500 then they are probably a spammer.

vaibhav-jain:
Hi @pydanny,

Could you please let me know which fields needs to be fixed, I checked some but they have `max-length=100` set

ryancheley:
It looks like there are 10 `TextFields` to review in the following Apps / Models:

## Grid

- Grid.description
- Feature.description
- Element.description

## Home Page

- PSA.body_text

## Package

- Category.description
- Package.repo_description
- Package.participants
- Package.commit_list

## SearchV2

- SearchV2.description
- SearchV2.participants

I don't think all of them need to have a limit of 500 characters, just those that are available from the front end. 

The next thing to do is:

- [ ] review the views for each model to see which fields are being called into the views
- [ ] review the templates to see what is being displayed on the front end

Once that is completed, a better assessment can be made (I think) about which fields should have the 500 character limit on them.

ryancheley:
The `participation` fields appear to both be pulled in from GitHub so I don't think they would need to have the 500 character limit since there may be (and in many large project cases probably are) cases where a 500 character limit would truncate the list of contributors which wouldn't be correct.

ryancheley:
Similarly, the `repo_description` and `commit_list` are also pulled from GitHub

ryancheley:
Fields not available from the front end:

- Category.description
- SearchV2.description
- PSA.body_text

Fields that are available in front end:

- Feature.description
- Grid.description
- ElementForm

PR #780, opened 2022-02-13
Limit text field size to 500 #111

ryancheley:
I added `max_length` to TextFields in the following apps / models:

# Grid

- Grid.description
- Feature.description
- Element.text

There were 7 other TextField in the project, but if you see my notes  ([here](https://github.com/djangopackages/djangopackages/issues/111#issuecomment-1038394539), [here](https://github.com/djangopackages/djangopackages/issues/111#issuecomment-1038398553), and [here](https://github.com/djangopackages/djangopackages/issues/111#issuecomment-1038405118)) I don't think they need to have the same limit (and in some cases may actually be detrimental)

# Homepage

- PSA.body_text

# Package

- Category.description
- Package.repo_description
- Package.participants
- Package.commit_list 

# SearchV2

- SearchV2.description
- SearchV2.participants

You'll notice there are a few migrations in this PR as I had initially applied the idea to all TextField but after the analysis done in the notes on #111 I thought maybe it should be more strategic.

Also, when I run the tests locally (just from main) I have a failing test `test_score`

This test also fails on my branch. Not sure if I set something up incorrectly, or ... 🤷‍♂️


jefftriplett:
@ryancheley thank you for the contribution. `test_score` has been giving me fits off and on as well. I have yet to nail down why, but I started some boring test changes tonight in an attempt to wrap my head around some of the test data. 

I'll think on the 500 limited, but it might take me a bit longer to verify what is going to change (get clipped) since this is a TextField. I'll do some more ticket cleanup too.

dmoney:
@ryancheley `test_score` should be fixed (#792).

jefftriplett:
Hey @ryancheley , thank you for your patience. It took me a while to get some other updates off of my plate to give the time it needed. 

I created a management command in #798 to help audit the TextFields and wrap my head around the domain. These are the results from production:

```shell
$ python manage.py audit_textfield_max_length
Elements: 6
pk=17735 :: 591
pk=4568 :: 901
pk=2531 :: 548
pk=2386 :: 946
pk=1847 :: 757
pk=1846 :: 543
Feature: 2
pk=1027 :: 501
pk=845 :: 582
Grid: 1
pk=136 :: 559 :: /grids/g/workers-queues-tasks/
PSA: 2
pk=8 :: 664
pk=7 :: 736
Package (repo_description): 1
pk=762 :: 504 :: /packages/p/django-pulse/
```

Ironically the grids that use the most fields are primarily from our CMS page  (see "Main focus" https://djangopackages.org/grids/g/cms/), and we have some other issues that need to be addressed.

- [ ] I think your assessment is pretty much spot on, but I suspect we really only want to restrict the `Package.repo_description` field to 500 characters. 

- [ ] The three Grid models/fields could be changed to have a `max_length=1000` and accomplish the same goal without clipping off any existing data. 

Let me know if you are up for these changes, and I can promise a much quicker review time on my part the second time around.

ryancheley:
@jefftriplett I’m up for making the changes! Thanks!

ryancheley:
@jefftriplett I've updated by branch and attempted to run the tests but get 2 failures:

```
======================================================================
FAIL: test_score (app.package.tests.test_models.VersionTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/app/package/tests/test_models.py", line 32, in test_score
    assert False
AssertionError

======================================================================
FAIL: test_score_abandoned_package (app.package.tests.test_models.VersionTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/app/package/tests/test_models.py", line 42, in test_score_abandoned_package
    self.assertEqual(p.score, 0, p.score)
AssertionError: -100 != 0 : -100

----------------------------------------------------------------------
Ran 111 tests in 124.140s

FAILED (failures=2)
Destroying test database for alias 'default'...
```

However, if I do the following to `app.package.tests.test_models.VersionTests`

1. Comment out line 32 `assert False`
2. Comment out line 42 `self.assertEqual(p.score, 0, p.score)` and uncomment line 43 `self.assertEqual(p.score, -100, p.score)`

Then the tests pass. Not sure if I'm using the wrong branch or something else. 

I'm going to start working on the changes you mentioned above.

jefftriplett:
Let's see what GH does with them. In this case, I find calling it a success if GH CI passes. I feel more comfortable as we move moreover to pytest and pytest-style tests in trying to get better DB isolation.

jefftriplett:
@ryancheley thank you for your contribution. 

I made a few changes and GitHub forced me to open a new PR which is now #804 to consolidate the migrations (to avoid clipping copy to 500 then unclipping it) and then I ran Black on everything. 

I'll merge it once tests pass and we'll deploy it tonight.

PR #781, opened 2022-02-16
update doc references from dev.yml to docker-compose.dev.yml

dmoney:
Updated commands in documentation files (*.rst) where the command referred to "dev.yml", changing them to "docker-compose.dev.yml".

There were two places referring to this that I didn't change:

* One was in `CHANGELOG.md`, and refers to the name of an issue rather than syntax of a currently used command.
* One was in `.devcontainer/devcontainer.json`, and that seems to be a VS Code config file, that I'm leery of messing with because I'm not currently using VS Code and don't know what this might affect.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,7 +70,7 @@ simply run::
 
 To run tests, run::
 
-    docker-compose -f dev.yml run django python manage.py test
+    docker-compose -f docker-compose.dev.yml run django python manage.py test
 
 The Site
 --------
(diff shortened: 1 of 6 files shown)


dmoney:
This is supposed to be a pull request for issue #782.

jefftriplett (approved):

jefftriplett:
@dmoney thank you for the contribution.

PR #792, opened 2022-02-21
fixes test_score test failure

dmoney:
Fix for the test_score failure in #784.

The django-cms package in package/tests/data.py was getting the 30% penalty for not supporting python3. Added supports_python3 to the latest version.


jefftriplett:
@dmoney thank you for the fix. 

(I went down a slightly different rabbit hole in #791 which changed the outcome based on that Version insertion order too which I hadn't seen in a while.)

PR #793, opened 2022-02-25
Re enable bitbucket test and fix bitbucket repo handler (Fixes #462)

### LINKED ISSUE #462: Re-enable bitbucket test
Bitbucket repo tests are currently disabled (https://github.com/pydanny/djangopackages/commit/4ac30c5420404ae4f3aa6cac05cad161e43af0b6).

dmoney:
Changes:

* Uncommented and fixed bitbucket repo handler tests
* Updated Bitbucket repo handler to Bitbucket API 2.0
* Updated repo handler test to use a valid repo instead of one that was removed
* Added script to find Bitbucket repos that still exist and have forks, in case that one goes away too
* Uncommented GitHub repo handler tests
* Added note to install doc about GitHub API key (necessary to run GitHub tests)


jefftriplett:
I wonder what is up with:

```
18.65s call     package/tests/test_repos.py::TestGithubRepo::test_fetch_metadata
```

It's taking ~40 seconds for me locally. I wonder if we can easily mock this?

dmoney:
Maybe that's why it was commented out before. 😅 

Mocking would probably be preferable to someone running the tests, but wouldn't let us know about API changes. Maybe there's a better way to monitor for that though.

jefftriplett:
Good work btw. Marking it as slow with a pytest marker might be good enough for now. I need to switch the GitHub workflow and docs to mention that at some point. 

Ideally, the GitHub library needs either updated or switched out with something else. It's our only blocker for upgrading from Python 3.7. That has to happen at some point either way. I'm mentioning that since it might make mocking up the request more trouble than it's worth.

jefftriplett:
It took a while to find it, but it looks like Django has so many contributors plus the rate-limiting code was slowing everything down. Filing under TIL and a huge opportunity to refactor. 

https://github.com/djangopackages/djangopackages/blob/main/package/repos/github.py#L57-L59

PR #796, opened 2022-02-27
Refactor code

ShreehariVaasishta:
Refactored some code to simplify without readability taking a hit
* Also makes use of `f-string` and `walrus` operator


jefftriplett:
Thanks for the update!

Issue #96, opened 2011-09-10
Flag to request removal of a package from a grid


pydanny:
Users should be able to flag a package for removal from a grid.

ShreehariVaasishta:
@jefftriplett Can you please some more info on this?

jefftriplett:
I didn't realize we hadn't closed this yet. I was working on flagged packages this morning.

PR #830, opened 2022-06-19
Add flagged package

jantarktic:
A potential solution to #96 

Here I've added a Flag model and subsequent views with a flag form.

Users will have the ability to flag a package and give a reason for their flag. 
Superusers then have the ability to review the flag and approve it in which package.html will display a small flag on it.
Users will also be able to see if a flag is pending review or not.

This is my first time writing these kinds of tests and adding such a large contribution to open source so I'm more than happy to keep refining this feature and any feedback at all is so greatly appreciated!


jefftriplett:
Thank you for your contribution. I will check it out this week and reply back with any feedback.

package/models.py
```diff
@@ -470,6 +471,23 @@ def last_commit(self):
...
+    created = models.DateTimeField(auto_now_add=True)
+    approved_flag = models.BooleanField(default=False)
+
+    def approve(self):
+        self.approved_flag = True
+        self.save()
+
+    class Meta:
+        ordering = ["-created"]
+        UniqueConstraint(fields=["package", "user"], name="unique_flagged_package")
```
jefftriplett, line 487: I think this would need to be activated with `constraints = [UniqueConstraint(fields=["package", "user"], name="unique_flagged_package")]`.

jefftriplett (approved):

jefftriplett:
@jantarktic This is great and apologies for the delays. 

I made a few small changes to consolidate down the migrations, and move flags over to the sidebar. All-in-all, great job and we really appreciate the contribution!

jantarktic:
Thanks Jeff!!

PR #837, opened 2022-08-30
:rocket: Homepage Cache

jefftriplett:
The goal here is to cache a few common homepage bits that take longer than we'd like to load.


luzfcb:
@jefftriplett Have you tried https://github.com/Suor/django-cacheops?

Seems like a good cache solution to me.
From my tests, it can cache an entire QuerySet/QuerySets, and it is possible to automatically invalidate a specific item in the cached queryset instead invalidate all.

Issue #848, opened 2022-09-24
Add support for emojificate


ryancheley:
**Is your feature request related to a problem? Please describe.**
A [recent issue of Django News](https://django-news.com/issues/146) mentioned a project called `emojificate` which

>  improves web accessibility by turning text with emojis into text with accessible emojis

I believe adding this could help with packages that have emoji in their titles from an accessibility perspective

**Describe the solution you'd like**
Add support for emojificate to the templates that currently have the `emojify` template tag

**Describe alternatives you've considered**
None (other than not implementing it)

**Additional context**
To Be Added

jefftriplett:
Fixed in #849

PR #849, opened 2022-09-24
Add emojificate

ryancheley:
A potential solution to #848


ryancheley:
One thing I have noticed with this implementation is that the default size for the emoji is 72x72 which may be a bit too big. There doesn't seem to be a way to override the size though. 

The fix for this may be one of:

* submit a PR to the `emojificate` package to support multiple sizes
* accept the larger emoji as it exists now
* not implement the `emojificate` package at all

ryancheley:
Ugh ... it looks like the change to `requirements.txt` that was passed to CI didn't include the actual new package 🤦🏻‍♂️

I'll work on that tomorrow (September 25). My laptop is close to dying and I didn't bring a charger to the coffee shop

ryancheley:
> Ugh ... it looks like the change to `requirements.txt` that was passed to CI didn't include the actual new package 🤦🏻‍♂️
> 
> I'll work on that tomorrow (September 25). My laptop is close to dying and I didn't bring a charger to the coffee shop

Looks like I got to it sooner than I thought I would 😀

ryancheley:
Here's an example of the difference

## Current Display

Screen Shot 2022-09-24 at 6 43 17 PM

## With emojification

Screen Shot 2022-09-24 at 6 43 23 PM

jefftriplett:
@ryancheley thank you for the contribution. I think you literally read my mind on this. I almost started it on Friday, but I thought I'd ask @glasnt a few questions after DjangoCon Europe. 

I'll check this out this week and see what a good workaround for the HUGE emoji might be.

glasnt:
Emojificate directly uses twemoji, which hosts svg and 72x72. At the time I decided png was preferable due to the limited svg implementations across all browsers. 

The hope was to replicate Twitter Web's accessibility, but that has since changed (they currently use SVGs, and have no tooltips)

I think the simplest solution in this case would be to add a class to the img tag and allow web users to add their own CSS to adjust for their display size. One of the [dependent repos](https://github.com/glasnt/emojificate/network/dependents) https://cz.pycon.org/2018/ has the same display issues due to the lack of this. 

It might also be useful to add a flag for svg. I've had a long standing issue https://github.com/glasnt/emojificate/issues/4 to allow for other renderings, so it would be in the same vein. 

I'll see about adding the class flag and let y'all know when a new release is available.

jefftriplett:
@glasnt thank you for the update. If I'm reading the tea leaves, it sounds like the CSS class might be our best bet to control styles. A Django setting for this might not be a bad way to go. If the value is set, `EMOJIFICATE_CSS_CLASS` the value is passed into `class="{EMOJIFICATE_CSS_CLASS}"` or something? Either way, it's not a big deal if it's too big of an ask.

glasnt:
The quick fix would be just adding `class='emojificate'` to the output. 

The better solution would be to have Django settings for this and the filetype that's configurable in the settings, defaulting to "png" and "emojificate" for the type and class respectively.

Where the type would serve up: 

* https://twemoji.maxcdn.com/v/latest/svg/2728.svg, or
* https://twemoji.maxcdn.com/v/latest/72x72/2728.png (default). 

I'll see about implementing this and let you know when a new version of emojificate is ready

glasnt:
emojificate 0.6.0a1 should now accept `EMOJIFICATE_CSS_CLASS` and `EMOJIFICATE_FILETYPE` settings. Tested locally, would appreciate a second look before I bump from pre-release to real-release.

ryancheley:
I can take a look this weekend and let you know what I find.

> emojificate 0.6.0a1 should now accept `EMOJIFICATE_CSS_CLASS` and `EMOJIFICATE_FILETYPE` settings. Tested locally, would appreciate a second look before I bump from pre-release to real-release.

ryancheley:
OK ... not sure what happened, but the database I had for local development is just gone and I'm having the hardest time getting data into the database. I'm sure I'm just missing something obvious ... but I'm not seeing it 😩 

Will try again tomorrow.

ryancheley:
Got the data back for the test package.

ryancheley:
I've been able to get it to work locally! Once emojificate gets a new release I'll update the requirements with the released version and push again

ryancheley:
@jefftriplett I resolved the conflicts locally and pushed the changes up, but I'm not sure if that is having a weird / negative impact on any PRs/Issues that are open. If there's any weirdness let me know when you get a chance. Thanks!

ryancheley:
In reviewing the `requirements.in` file I realize I did NOT need to pin `emojificate` to a specific version. This was done initially for testing when the needed feature was in pre-release. I’ll be making an update and pushing later

jefftriplett:
@ryancheley if you don't mind pulling in my changes to update your fork, it should be good to go. The branch is called `add_emojificate` and I resolved the requirements.* changes.

ryancheley:
@jefftriplett fork has been updated. Thanks!

jefftriplett:
@ryancheley @glasnt thank you both!

PR #853, opened 2022-10-21
:gear: Adds justfile

jefftriplett:
Mostly there but kind of rough still. 

cc @ryancheley


ryancheley:
@jefftriplett Looks good to me

PR #857, opened 2022-10-27
Doc refactor 855

ryancheley:


ryancheley:
Initial draft for #857

jefftriplett:
Nice work. I'll pull and check it out later. Have any opinions about sphinx vs. MkDocs? I noticed that RTD supported both last night when I was unbreaking our doc builds.

ryancheley:
MkDocs looks interesting. And if it’s means we get to write the docs in Markdown instead of reStructuredText I’d be all for it 😁

jefftriplett (approved):

jefftriplett:
@ryancheley good work. I loosely looked (on a laptop vs. bigger screen) but I think every small change is a big win.

PR #858, opened 2022-10-27
Fix N+1 Query on Grid Detail Page [`package.category`]

saadmk11:
Ref:

https://github.com/djangopackages/djangopackages/blob/e4806a86a51c71f1a62f5c61d8353adc679f53fe/templates/grid/grid_detail.html#L142

### DIFF
--- a/grid/views.py
+++ b/grid/views.py
@@ -296,9 +296,9 @@ def grid_detail(request, slug, template_name="grid/grid_detail.html"):
 
     # features = grid.feature_set.select_related(None)
     features = Feature.objects.filter(grid=grid)
-    grid_packages = grid.grid_packages.select_related("package").filter(
-        package__score__gte=max(0, settings.PACKAGE_SCORE_MIN)
-    )
+    grid_packages = grid.grid_packages.select_related(
+        "package", "package__category"
+    ).filter(package__score__gte=max(0, settings.PACKAGE_SCORE_MIN))
 
     filter_form = GridPackageFilterForm(request.GET)
 


saadmk11:
I saw that the `grid_detail.html` template also calls `grid_package.package.last_released` twice for each package which causes `2N` queries if the **`cache` misses** or if the **package does not have any versions**.

https://github.com/djangopackages/djangopackages/blob/e4806a86a51c71f1a62f5c61d8353adc679f53fe/package/models.py#L438-L445

We can probably improve that by using `@cached_property` or fix it by using annotations:

```python
    versions = (
        Version.objects.filter(package_id=OuterRef("package_id"))
        .exclude(upload_time=None)
        .order_by("-upload_time")
    )
    grid_packages = (
        grid.grid_packages.select_related("package", "package__category")
        .annotate(
            supports_python3=Subquery(versions.values("supports_python3")[:1]),
            pretty_status=Subquery(versions.values("development_status")[:1]),
        )
        .filter(package__score__gte=max(0, settings.PACKAGE_SCORE_MIN))
    )
```
It may be worth it if this happens often in production. It would be great to hear your thoughts @jefftriplett :)

jefftriplett:
Nice catch. I'll pull this later and test it. 

I meant to add some additional testing to check assertNumQueries() but I haven't quite made it around to the grids yet. 

side-note: I have been tempted to pull more of the caching code out because it seems to be a sore spot for bugs and performance issues.

jefftriplett (approved):

jefftriplett:
Thanks again!

saadmk11:
> Nice catch. I'll pull this later and test it.
> 
> I meant to add some additional testing to check assertNumQueries() but I haven't quite made it around to the grids yet.
> 
> side-note: I have been tempted to pull more of the caching code out because it seems to be a sore spot for bugs and performance issues.

Awesome, Let me know if I can help with any of that. I am happy to contribute. :)

Issue #855, opened 2022-10-25
:books: Refactor/refresh/rewrite the contribution docs


ryancheley:
**Is your feature request related to a problem? Please describe.**
I'd like to Refactor/refresh/rewrite contribution docs to make it easier for for people to walk through the initial setup and kick off process.

**Describe the solution you'd like**
Review the docs and update:

1. Update the `justfile` to include comments
2. Update the `faq.rst` to have a link to the contribution guide
3. Update the `contributing.rst` to include steps on using `just` and the `justfile`
4. Review the `readme.rst` for accuracy and update as needed

There may be other items discovered during the refactor phase. 

**Additional context**
This idea was based on [this conversation](https://twitter.com/webology/status/1583837021902028800?s=20&t=K6vQD38FEXwCaiBtBsRFzA) on Twitter

I'm more than happy to take this task on!

ryancheley:
In reviewing the `justfile` there are a couple of commands that I'm still researching:

- caddy-fmt
- caddy-validate
- lint-fmt
- restore

The `caddy` options require `caddy` to be installed. I'm not sure if this is a production only thing or something else

The `lint-fmt` requires additional dependencies (`black`, `tryceratops`, `djhtml`) which one may want to install with `pipx`. Alternatively you could have a virtual environment set up locally, but in either case, the `readme.rst` doesn't have any reference to that (yet!)

The `restore` seem incomplete. The `docker-compose` commands are all commented out

ryancheley:
@jefftriplett I'm reviewing the docs and have a few questions / am seeking input on. There are a few questions and it may be best to move this to a discussion instead of inline on the issue. If a discussion is a better spot for this, let me know and I'll create a discussion and move the questions there. 

# Questions

1. There are several links to ReadTheDocs, but the ReadTheDocs doesn't seem to match what's in the repo? Should we
   a. remove the references to the ReadTheDocs and drop that site?
   b. determine how to build the docs on ReadTheDocs and incorporate that into the publish process?
2. There are tips on the [Contributing page ](https://github.com/djangopackages/djangopackages/blob/main/docs/contributing.rst#tips) that seem like they may no longer apply with respect to the `starter` label and the `doc` label. I think the tips section should be removed
3. The docs reference the `develop` branch to be used for fetching upstream changes, but there isn't a develop branch. Should the docs be updated, or does a new branch need to be added

jefftriplett:
[...]

PR #861, opened 2022-10-28
Typos fixes

ryancheley:
There were a few typos that prevented the content from being displayed correctly; also made sure that `just` was used consistently.

Refs #855

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -36,6 +36,7 @@ You'll want to install the various formatters, and linters using the following `
 To do this, run:
 
 .. code-block:: bash
+
     just pipx-install
 
 Grab a Local Copy of the Project
@@ -56,6 +57,7 @@ Set Up Your Development Environment
 In order to run the project, you'll need to run the following command:
 
 .. code-block:: bash
+
     just setup
 
 Add A GitHub API Token
@@ -72,7 +74,7 @@ Now build the project using docker-compose:
 
 .. code-block:: bash
 
-    docker-compose build
+    just build
 
 Run the Project
 ---------------
@@ -81,7 +83,7 @@ To start the project, run:
 
 .. code-block:: bash
 
-    docker-compose up
+    just up
 
 Then point your browser to http://localhost:8000 and start hacking!
 
@@ -94,7 +96,7 @@ Create a Django superuser for yourself, replacing joe with your username/email:
 
 .. code-block:: bash
 
-    docker-compose run django python manage.py createsuperuser --username=joe --email=joe@example.com
+    just superuser joe joe@example.com
 
 And then login into the admin interface (/admin/) and create a profile for your user filling all the fields with any data.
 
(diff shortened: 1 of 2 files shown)


jefftriplett:
`just` is probably so new/different to people that I worry it might be a barrier vs. giving people compose commands by default. I don't mind mentioning a just workflow too, but if that complicates things then we can skip it in favor of maybe doing a "how we work page" section or page or something. 

Thoughts on this? Just thinking of small stones in the road people might go over and not notice or they might stumble and then it adds another layer on top of a few layers?

ryancheley:
🤦🏻‍♂️ I’m just a fan of just I assume that **everyone** else is also … doh!

I like the idea of a ‘how we work’ section. I can go back and swap out the `just` commands with their `docker-compose` versions and then create a new page/section that would include `just` commands with an intro about what `just` is and why we use it and why someone else might want to use it as well. 

This would, I think, allow someone not familiar with just to get started but also:

1. Allow people that are familiar with just to use it
2. perhaps pique the curiosity of people not familiar with just to ‘give it a try’

I think we may also want to update the `docker-compose` commands to `docker compose` (no dash) which is compose version 2.

It was [put into GA in April 2022](https://www.docker.com/blog/announcing-compose-v2-general-availability/), though I’m not sure how wide spread it’s use is. 

Maybe the best way to approach **that** is to keep `docker-compose` but have a note somewhere that `docker compose` will work if you’re on compose version 2. Any thoughts on that? 

I’ll start making some updates and submit another PR with them.

ryancheley:
@jefftriplett I split out an 'opinionated' install set of instructions and then added `docker-compose` commands with references to the use of `just`. Take a look and let me know what you think.

docs/install.rst
```diff
@@ -81,7 +83,7 @@ To start the project, run:
 
 .. code-block:: bash
 
-    docker-compose up
+    docker-compose --file docker-compose.dev.yml up --detach
```
jefftriplett, line 86:
```suggestion
    docker-compose up --detach
```

docs/install.rst
```diff
@@ -72,7 +74,7 @@ Now build the project using docker-compose:
 
 .. code-block:: bash
 
-    docker-compose build
+    docker-compose  --file docker-compose.dev.yml build
```
jefftriplett, line 77:
```suggestion
    docker-compose build
```

jefftriplett:
I removed the `docker-compose.dev.yml` flow in an attempt to have one less abstraction for people to work with. I left a few suggestions but that's really the bulk of what needs changed to make local dev easier. 

I will probably open an issue on using pre-commit and that should clean up my usage `pipx ..` a bunch of things which I know is kind of annoying. This is a me problem though and I can fix it. 

That said, I'm away to see family this weekend, but I'll try and slip some time in either Friday, Sunday night, or next week to give it a good checkout and review.

ryancheley:
The changes all made sense to me. I like the idea of adding pre-commit to the project to help new contributors. I've been working on a bug in [django](https://github.com/django/django/pull/16243) and I've submitted my code (more than once!) without linting it to only have the linting tests not pass and I'm like ... not again! 😩

jefftriplett:
Merged, and we can sort out iterative changes as we go. Thanks again.

PR #870, opened 2022-11-07
:fire: Removes `--file docker-compose.dev.yml` refs

jefftriplett:


.devcontainer/devcontainer.json
```diff
@@ -3,7 +3,6 @@
 {
     "name": "Existing Docker Compose (Extend)",
     "dockerComposeFile": [
-        "../docker-compose.dev.yml",
```
smithdc1, line 6: Does this not need to refer to `docker-compose.yml`?

I made that edit the weekend and managed to get it running for me locally. But I could well be doing it wrong as I've never used dev containers before.

PR #872, opened 2022-11-12
✨ Add widont support

jefftriplett:
Adds "widont" support to avoid hanging words on the homepage, which is where it annoys me the most. 

The `django-typogrify` is currently not supported for Django 4.0, so I pulled in just what we needed. It has some other useful tags in it though too.


jefftriplett:
/me deploys to scratch and itch, but cc'ing everyone for awareness.

PR #873, opened 2022-11-13
📚  Minor doc fixes

jefftriplett:
I noticed these were warning/erroring tonight so here are a few updates to address a few.

### DIFF
--- a/docs/deployment/docker.rst
+++ b/docs/deployment/docker.rst
@@ -16,8 +16,7 @@ restarded, the stack won't start automatically.
 
 The current strategy is:
 
- - Use a virtual machine with a well patched OS (debian, ubuntu, RHEL), djangopackages.org is using
- ubuntu 14.04
+ - Use a virtual machine with a well patched OS (debian, ubuntu, RHEL), djangopackages.org is using ubuntu 14.04
  - Install docker, docker-compose, git and supervisord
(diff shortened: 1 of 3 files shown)


docs/deployment/docker.rst
```diff
@@ -16,8 +16,7 @@ restarded, the stack won't start automatically.
 
 The current strategy is:
 
- - Use a virtual machine with a well patched OS (debian, ubuntu, RHEL), djangopackages.org is using
- ubuntu 14.04
+ - Use a virtual machine with a well patched OS (debian, ubuntu, RHEL), djangopackages.org is using ubuntu 14.04
```
ryancheley, line 19: @jefftriplett Is the server still using Ubuntu 14.04?

jefftriplett: It's still on the OG sever from ~8 or ~9 years ago AFAICT. I'd like to move it over to a new slice over one of our holiday breaks over the next month or so. It shouldn't be too bad since I moved the DB out, and it's mostly container based outside of cron-jobs. I'll open a ticket fo this.

jefftriplett:
I'm going to close this since #874 covers a few things I was checking off the list. I need to become more familiar with Sphinx and probably print the docs to get a better idea of what they should even look like.

Issue #882, opened 2022-11-25
"Sites using this package" contains suspicious links 👀


jefftriplett:
> also I noticed that the section "Sites using this package" contains suspicious links 👀 #881 

- [x] We have 1518 package examples, and I suspect most are spam. A reasonable solution for this might be to change the default PackageExample.active to `None`. Fixed in #883 

I'm not sure how to best deal with the existing packages, but if there is a reasonable "check this domain for spam" service, it might be worth a scheduled script to detect them. (mostly thinking out loud about this. We could remove them entirely, and I'm not sure we'd be out anything)

jefftriplett:
I have been updating these by hand. Only ~5 over the last month and a half, so it's manageable.

PR #883, opened 2022-11-25
:elephant: Default PackageExample.active to None

jefftriplett:
Follow up from #882 

This PR:
- [x] changes the default value of `PackageExample.active` from `True` to `None` to avoid spam showing up by default on the website. That way a `None` value means a `PackageExample` record is new and a `True` can represent approved, and a `False` would mean not approved. 
- [x] updates the package detail template to add `rel="nofollow"` tags to each link. 
- [x] adds three admin actions to make bulk updating a little easier.


jefftriplett:
Merging but happy to take post-merge review feedback.

ryancheley (commented):
Looks good to me.

PR #885, opened 2022-11-27
:sparkles: Adds check_package_examples

jefftriplett:
Addresses part of #882


jefftriplett:
Merging to move this to production to try out.

PR #890, opened 2022-11-30
:tractor: Refactors the grid detail layout section ordering

jefftriplett:
WIP updates the ordering of sections on the grid details page. I know originally it was a design decision to add the CRUD elements to the top, but Grids don't change as much these days, and it's a big scroll to get there. 

It also adds `emojificate` support for the grids so that titles and descriptions get emoji coverage.

I'd like to change up the filter results and sort by to be an inline style so that we don't have as much whitespace. 

screencapture-localhost-8000-grids-g-database-migration-2022-11-30-14_21_15


ryancheley:
I think this is a really good idea. Are there any metrics that show IF users access the grids from the top level navigation. It probably won't matter (too) much, but there could be some users that expect to see the top level grid for **reasons**.

jefftriplett:
@ryancheley, by top-level grid do you mean the features overview? 

Eventually, that might be nicer to be a hover dialog or hover text over the grid columns too. I think it's skewed on longer grids now and the feature grids haven't been maintained in several years. Might be worth noting there isn't enough data to render them as a motivator for people to help fill them in. 🤔

ryancheley:
@jefftriplett 

> by top-level grid do you mean the features overview?

When I'm running the change locally and looking at [/grids/](http://localhost:8000/grids/) (or any page really) this is what I'm seeing:

local header

versus what's on the [live site](https://djangopackages.org/grids/):

prod header

jefftriplett:
@ryancheley Ahh, that makes more sense. I don't think you can see the "Add Grid" button unless you are logged in.

jefftriplett:
I opened this for review since the only red flag seemed to be unrelated. I do plan to flatten the filters, but I'm sick at home and probably won't get to that sooner than later.

ryancheley:
@jefftriplett OK ... finally got my head wrapped around this. Some differences on my local environment made it so I couldn't see what was going on 🤦 (specifically with the top level navigation)

Just to confirm, the intended changes:

1. Move the `Add package to the grid` and `Add new feature to the grid` buttons to the bottom of the page instead of the top of the page

local footer

2. Add the grid description to the page (below the Grid Name)

If not, then I'm probably missing good test data to locally to see what the overall intention is in runtime

jefftriplett:
Check out the Calendar grid https://djangopackages.org/grids/g/calendar/ which is a fairly common example of features gone bad. 

Xnapper-2022-12-06-13 51 31-AC3187D5-3815-41AA-A208-6819DEC78F38

Notice the features + filter results push the actual grids a full screen (aka below the fold) even to see them. 

I'm open to moving the "Add ..." button links to where they make sense.

ryancheley:
OK. That really helped. I think this does look much better and puts the actual content that we would want to see closer to the top where it goes! 👍

jefftriplett:
@ryancheley thanks. Merging, and we can refine.

PR #895, opened 2023-01-04
📈  Funding updates

jefftriplett:
This removes our `FUNDING.yml` page, and it's something we can address/figure out an org at some point.

I added a few charts to /open/ with the intent to start documenting monthly/annual expenses for running the website. I have hosting, domain registration, and ethical adds income listed. Ideally, this could be a Django model and be expanded.

### DIFF
--- a/templates/pages/open.html
+++ b/templates/pages/open.html
@@ -115,7 +115,7 @@
                 </table>
             </div>
             <div class="col-sm-6 col-md-5">
-                <h2>{% trans "Grids with twenty or more packages" %}</h2>
+                <h2>{% trans "Grids with twenty-five or more packages" %}</h2>
                 <table class="table">
                     <thead>
                         <tr>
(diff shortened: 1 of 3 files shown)


templates/pages/open.html
```diff
@@ -89,7 +115,7 @@ <h2>{% trans "Users who have contributed over ten packages" %}</h2>
                 </table>
             </div>
             <div class="col-sm-6 col-md-5">
-                <h2>{% trans "Grids with more than fifteen packages" %}</h2>
+                <h2>{% trans "Grids with twenty or more packages" %}</h2>
```
ryancheley, line 118: I noticed that the code was updated to ‘ num_packages__gte=25’ but the copy displayed on the page says twenty or more. Not sure if that **really** matters or not

jefftriplett, line 118:
```suggestion
                <h2>{% trans "Grids with twenty-five or more packages" %}</h2>
```

jefftriplett: Nice catch. Thank you. I was visually making changes and fiddling, and that one got away from me.

jefftriplett:
@mergifyio refresh

PR #896, opened 2023-01-04
:sparkles: Adds Classifiers and Products apps

jefftriplett:
The Classifiers app allows us to track PyPI's [trove classifiers](https://pypi.org/classifiers/) and periodically update them. 

The Products app tracks Release, End-of-life (EOL), and other data for Django, Python, Wagtail, and future projects using the https://endoflife.date/ API.

Both of these apps add some useful metadata that we might have to otherwise hardcode. My end goal is to build some better "what is relevant" features for search and grids.


ryancheley:
@jefftriplett will the management commands be run on an ad hoc basis or if they’ll be part of a cron job of some kind? I haven’t had a chance to download the updated code and run locally, but am hoping to do that this weekend (first week back at work has been challenging!)

jefftriplett:
> @jefftriplett will the management commands be run on an ad hoc basis or if they’ll be part of a cron job of some kind? I haven’t had a chance to download the updated code and run locally, but am hoping to do that this weekend (first week back at work has been challenging!)

Good questions, and this is a good nudge for me to update docs. They'll run via cron, but post-server move, I might move them over to celery since that's more portable IMO.

I go back next week so I'm hoping it's uneventful but I get the newly back-to-work blues.

classifiers/models.py
```diff
@@ -0,0 +1,21 @@
...
+from django.utils.translation import gettext_lazy as _
+from django_better_admin_arrayfield.models.fields import ArrayField
+
+from core.models import BaseModel
+
+
+class Classifier(BaseModel):
+    classifier = models.TextField()
+    active = models.BooleanField(
+        _("Active"),
```
ryancheley, line 11: I see what this line is doing, but I'm not sure WHY it's doing it. I can't find anything in the docs that tell me why this works, or which **option it's setting the "Active" for. Any links you have to help point me in the right direction?

ryancheley (commented):
I know this has already been merged, but I did want to go through it and try to understand the updates that were made. I did have one question on the use of BooleanField which is just for my own education.

classifiers/models.py
jefftriplett: Do you mean the `_("Active")` field label itself? 

Assuming so, you can override the label in the admin and model forms. They _() is a shortcut to get language translations. I think I copied the field from `package.models` when I was looking for prior art. Otherwise, I might have dropped the translations since we aren't taking advantage of them. I'm not against addressing that though.

ryancheley: > Do you mean the `_("Active")` field label itself?

Yes, that’s what I was asking about 😁

I didn’t notice the `from django.utils.translation import gettext_lazy as _ ` and now it totally makes sense! 

I’ve seen that used in [Enumeration types](https://docs.djangoproject.com/en/4.1/ref/models/fields/#enumeration-types) but not on other field types (like BooleanField) so it was new to me and I didn’t realize it could be done in other places. 

> Otherwise, I might have dropped the translations since we aren't taking advantage of them. I'm not against addressing that though.

In terms of addressing do you mean to drop the option for translations OR to set up translations? 

There don’t appear to be any issues (open or closed) asking for translations so I’m not sure if that means no one needs/wants them OR non-English speakers don’t know about the site and therefore haven’t had the opportunity to ask!

jefftriplett: > In terms of addressing do you mean to drop the option for translations OR to set up translations?

> There don’t appear to be any issues (open or closed) asking for translations so I’m not sure if that means no one needs/wants them OR non-English speakers don’t know about the site and therefore haven’t had the opportunity to ask!

I suspect translations via Python and the Django templates were added as a best practice or some future endeavor. I didn't see any tickets/open issues for this either when I started working on the project ~2 years ago.

I guess I feel like, let's do it right or not do them at all, where I lean towards "let's do it right". However, I really want to fix grids, search, and figure out how to remove more spam since those impact the overall relevance of the project. So maybe a good 2024 goal if we can land in a good place this year.

ryancheley: Agreed! I’ll write up an issue so that we don’t lose track of it.

Issue #644, opened 2021-08-06
✨ Update Open Graph tags and add per page preview images


jefftriplett:
The bulk of this is fixed in #869 

I'm going to leave this open until we can verify we land preview image and the template.

saadmk11:
If we want per-page image preview, I built a prototype using `Pillow` (Background is a static image created using Figma ).

![output](https://user-images.githubusercontent.com/24854406/205067129-a5128549-9266-4030-ac83-df99f93ab4b6.png)

**Note:** We can add more data depending on requirements.

We can follow 2 approaches to implement this:

1. **Image Field in Model: (Recommended)**
    - We need to add an image field on the Packages model
    - On Package create we generate the Preview image
    - We re-generate on every save action of the package model or use a celery task / cron job to update the images periodically

2. **Cached Django View:**
    - We create a cached Django view that returns image response
    - We cache the view for 24h
    - We do not need to run cron job / celery task periodically

jefftriplett:
For a few projects, I have written a few HTML templates which fit the open graph preview images and created a cached view form them. The only annoyance has been having to either celery update the images or use an HTML to PNG proxy which there are dozens of. 

Personally, I think an HTML template is easier to maintain, but I have never quite figured out a good system for taking screenshots that I could run in Docker well. 

For DjangoCon US this year we used a combo of [shot-scraper](https://github.com/djangocon/2022.djangocon.us/blob/main/justfile#L26) and [11ty's screenshot API](https://www.11ty.dev/docs/services/screenshots/). See this [example](https://github.com/djangocon/2022.djangocon.us/blob/abdf8e369590511daac3ab46e9271c2d7380da8f/_schedule/talks/2022-10-18-17-10-t0-tips-and-tricks-for-optimizing-django.md?plain=1#L16)

I think I have a half dozen attempts at this locally, and the blocker of getting Playwright to run on top of the base Python Docker Image has always tripped me up.

I'm definitely open to ideas (Pillow, etc) because should be a solved problem by now.

saadmk11:
Yes, I have tried several of these before but none of the quite fit right and some of them require bulky dependencies and a lot of workarounds to run on Docker. So, this time I tried to use `Pillow`. It was fairly simple to write/draw on top of a background image. I also understand your concerns about maintainability.

jefftriplett:
[...]

PR #902, opened 2023-01-30
:sparkles: Adds opengraph templates and views for Packages and Grids

jefftriplett:
More supporting work for #644


jefftriplett:
Skipping normal reviews so I can do some work that needs this.

PR #905, opened 2023-02-03
:pencil: Doc and justfile updates

jefftriplett:
I did some light updates earlier and updated our justfile to start documenting some better workflows.


jefftriplett:
I'm going to try to spend some time rewriting parts of the docs to make onboarding new users easier. I have had a few people get lost along the way, so I'm going merge and do some follow up.

justfile
```diff
@@ -50,40 +60,122 @@ bootstrap *ARGS:
...
+# --------------------------------------------------
+# Caddy recipes
+# --------------------------------------------------
+
+# Format our Caddyfile
 @caddy-fmt:
     docker-compose run --rm caddy caddy fmt -overwrite /etc/caddy/Caddyfile
 
-# ???
+# Ir our Caddyfile valid?
```
ryancheley, line 79: @jefftriplett looks like there's a typo here

`Ir our Caddyfile valid?`

Should be

`Is our Caddyfile valid?`

ryancheley: This should be fixed in #907

jefftriplett: 🤦  good catch. I have a huge stash that I trying not to lose while doing this.

PR #908, opened 2023-02-07
Reduces UX copy to reduce cognitive load.

daheats:


jefftriplett (approved):

jefftriplett:
Thank you for the PR. We'll probably revisit and pull the story copy into its own page since there is history with the origin story. Just noting that here since we'll probably focus on docs this month.

PR #910, opened 2023-02-10
:sparkles: 2023-02-10 Doc updates

jefftriplett:
Trying to feel out doc layout a bit and starting to document management commands.


jefftriplett:
cc @williln for some info

docs/install.md
```diff
@@ -48,29 +27,31 @@ cd djangopackages
 
 In order to run the project, you'll need to run the following command:
 
-```bash
-docker-compose build --force-rm
-```
-
-### Add A GitHub API Token
+All of the environment variables and settings that are needed to run the project are stored in  `.env.local.example` file.
```
williln, line 30: @jefftriplett I would move this sentence to before "In order to run the project..." on line 28. That way, your statement, "you'll need to run the following command..." is immediately followed by the command, so the flow of instructions keeps going.

docs/management_commands.md
```diff
@@ -1,83 +1,100 @@
...
+
+## pypi_updater
 
 To update packages with the latest data on PyPi, run:
 
-```bash
+```shell
 docker-compose run django python manage.py pypi_updater
 ```
+Warning: This can take a long, long time.
```
williln, line 86: This is an excellent warning to add!

williln:
Looks great to me @jefftriplett! I left a suggestion but it looks great. It's nice to see the commands in alpha order.

docs/install.md
jefftriplett: danke

.env.local.example
```diff
@@ -1,14 +1,10 @@
...
-DEBUG=
-DJANGO_DEBUG=
+DJANGO_DEBUG=True
 DJANGO_MAILGUN_API_KEY=
 DJANGO_SENTRY_DSN=
-DJANGO_SETTINGS_MODULE=
-DOMAIN_NAME=
-EMAIL_HOST_USER=
-FRAMEWORK_TITLE=
+DJANGO_SETTINGS_MODULE=settings.docker
```
williln, line 7: Was literally about to DM you about this!

.env.local.example
```diff
@@ -1,14 +1,10 @@
...
 DATABASE_URL=postgres://djangopackages:djangopackages@postgres:5432/djangopackages
-DEBUG=
-DJANGO_DEBUG=
+DJANGO_DEBUG=True
 DJANGO_MAILGUN_API_KEY=
 DJANGO_SENTRY_DSN=
-DJANGO_SETTINGS_MODULE=
-DOMAIN_NAME=
-EMAIL_HOST_USER=
-FRAMEWORK_TITLE=
```
williln, line 11: FWIW I did need `FRAMEWORK_TITLE` to make it run locally. I got an error without it. (I copied and pasted your new `.env.local.example`)

jefftriplett: Thank you for this. Maybe it's time for "one settings file to rule them all."

.env.local.example
jefftriplett: Follow up in https://github.com/djangopackages/djangopackages/issues/915

PR #914, opened 2023-02-11
Docs: Refresh the index page 

williln:
**NOTE**: I do not know how to run the docs locally, so I couldn't check if this all rendered accurately. But I would love to figure out how to run the docs locally and add that info to the docs 😄 

## Changes in this PR: 

- Adds a new intro to the index page 
- Adds a section to the index page with a brief explanation of the main concepts of the project (took a stab at defining the terms myself -- feedback welcome) 
- Moves the history of the project to a `history.md` file and adds that to the table of contents on the index page 
- Removes the "Lessons Learned" page and adds most of that info to the new `history.md` page. Also removes the lessons learned from the table of contents on the index page 
- Removes the "introduction" page since I moved most of the other content to other pages 

cc @jefftriplett


jefftriplett:
@williln 🎉  Thank you. I'll work out some docs instructions too.

PR #918, opened 2023-02-11
:tractor: Updates docs domain

jefftriplett:
This moves our default docs home/location to https://docs.djangopackages.org/en/latest/


jefftriplett:
cc @williln @ryancheley

PR #922, opened 2023-02-11
:tractor: Adds pre-commit

jefftriplett:
This needs a doc update and to firm up which rules we use and don't since we didn't quite get 100% of our `pipx` utils integrated. 

I don't really run pre-commit as a git hook, but that seems to be the default/normal way of using it. 

What this needs still:
- [x] pre-commit CI check
- [ ] update docs
- [ ] audit pipx list from docs and pull in anything we are missing unless ruff already covers it. 

Also, I still struggle a bit with if ruff covers all of black or if both are still needed.


.pre-commit-config.yaml
```diff
@@ -43,50 +47,35 @@ repos:
...
-      - id: flake8
-        additional_dependencies:
-          - flake8-bugbear
-          - flake8-comprehensions
-          - flake8-print
-          - flake8-no-pep420
-          - flake8-tidy-imports
-  - repo: https://github.com/codespell-project/codespell
-    rev: v2.2.2
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
```
pauloxnet, line 71: I think it's great that you use `ruff`, I've also started using it instead of `flake8` and `isort`, but it also replaces `pyupgrade` which you can remove.
https://github.com/charliermarsh/ruff#pyupgrade-up

jefftriplett: @pauloxnet thank you for the tip. I struggle with what ruff can lint vs. replace (like black{ish} support). Not sure if you have any public configs yet, but this is going to prompt me to check out some of your public repos again.

jefftriplett:
pre-commit.ci is installed now. Merging and sorting out docs next.

PR #929, opened 2023-02-18
:sparkles: Adds very alpha Django/Python/Wagtail Readiness

### LINKED ISSUE #741: ✨ pyreadiness but for django
Someone pinged about having a page like the PyReadiness project, which shows the top 250 Python packages and whether they support a given version of Python. See: https://pyreadiness.org/3.10/

I think we already have the data with our recent model changes to pull this off for Django. Some of this package metadata might be a good fit for the package_detail page. 

We'd want to look into the `pypi_classifiers` field and sort by `pypi_downloads` (I think `pypi_downloads` needs to be fixed via #834 ) to give a good feel of the top-250 packages. If we don't have the data in `pypi_downloads` we could use `repo_watchers` instead. 

https://github.com/djangopackages/djangopackages/blob/main/package/models.py#L66-L71

I think this discussion is worth looking into before we move forward: https://github.com/di/pyreadiness/issues/18

jefftriplett:
Fixes #741 

Small preview. It needs the legend spelled out and more data to be useful. 

This checks a few boxes but it'll be a bit before I link it anywhere. 

Xnapper-2023-02-18-14 27 50

This feature also sneaks in Tailwind CSS into our stack.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -4,10 +4,12 @@ import logging
 import pytest
 from django.test.utils import override_settings
 
+
 pytest_plugins = [
     "grid.tests.fixtures",
     "homepage.tests.fixtures",
     "package.tests.fixtures",
+    "products.tests.fixtures",
 ]
 
 
--- a/homepage/tests/test_views.py
+++ b/homepage/tests/test_views.py
@@ -1,3 +1,7 @@
+import pytest
+
+from django.urls.exceptions import NoReverseMatch
+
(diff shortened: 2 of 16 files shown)


jefftriplett:
I merged, but this is not linked up yet. We don't quite have data feeding in consistently yet for download numbers and some other stats. Calling this a mental win and working on some data issues.

PR #934, opened 2023-02-26
Enable `django-debug-toolbar`

saadmk11:
This PR:

1. Enables `django-debug-toolbar` for development
2. Fixes Test Failure when `django-debug-toolbar` is enabled.

Alternative:

- Using [pytest-is-running](https://github.com/adamchainz/pytest-is-running) package.

Discussion:
- https://github.com/pytest-dev/pytest/issues/9502

### DIFF
--- a/settings/base.py
+++ b/settings/base.py
@@ -1,5 +1,5 @@
 # Django settings
-
+import sys
 from pathlib import Path
 
 import environ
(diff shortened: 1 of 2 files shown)


jefftriplett (approved):
🎉  looks great!

PR #955, opened 2023-03-18
Decrease Homepage Database Query Count

saadmk11:


jefftriplett:
I'm kind of torn on this because I'd like to bring back grid-of-the-week and the PSA (which I can't find the template for), but I appreciate removing four queries from the homepage. I'll sleep on this, but probably create two new tickets to bring those back at a later date.

ryancheley:
@jefftriplett I think it’s a good idea to remove the queries for now and have separate issues to reintroduce them at a later date. This allows the home page to load more quickly now (win!) and to not forget that these things are important and can/should be introduced back at a later (another win!). Also, introducing them as separate issues allows (potential) discussion about them and we may discover things about those features we didn’t even know!

jefftriplett:
@ryancheley solid points.

PR #998, opened 2023-07-21
Add Carbon Ads

sayzlim:
Install Carbon Ads on the site. Each page will have a different zone key to track the impressions and performance separately. Some of the containers have been set to `col-lg-12` to make sure they look okay on the mobile devices.

@jefftriplett can you run it locally to see if there are any issues with the layout? I can move the ad and update the CSS if you have any preferences.

### DIFF
--- a/templates/grid/grid_detail.html
+++ b/templates/grid/grid_detail.html
@@ -15,12 +15,11 @@
 
 {% block extra_head %}
     {{ block.super }}
-    {% include 'includes/_ethicalads-js.html' %}
 {% endblock %}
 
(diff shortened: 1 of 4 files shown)


jefftriplett:
@sayzlim thank you for the PR. Apologies for the delay. I tend to block out my Friday afternoons to work on open-source projects, but we had a family trip last week, and everything snowballed. I'll be checking this out tonight and tomorrow.

jefftriplett:
I accidentally merged this, but sorting a few things out and verifying locally first before I deploy it.

Issue #961, opened 2023-03-27
:whale: Add a `worker` service


jefftriplett:
We loosely have celery support but have never switched over to it. The way cronjobs run today, they can take up the `django-a` process, so a nice stopgap until we support celery would be to add `worker` service and re-point our cronjobs to it for running our daily cron jobs. 

The worker service does not need to auto-run or restart. It should only be used for cronjobs without any external port mapping. 

```shell
12 23 * * * /usr/bin/docker compose -f /code/djangopackages/docker-compose.prod.yml run --rm django-a python manage.py searchv2_build
36 17 * * * /usr/bin/docker compose -f /code/djangopackages/docker-compose.prod.yml run --rm django-a python manage.py package_updater
48 21 * * * /usr/bin/docker compose -f /code/djangopackages/docker-compose.prod.yml run --rm django-a python manage.py pypi_updater
```

jefftriplett:
I'm going to bump this up in priority because I think it might be bubbling up to impact the production service.

saadmk11:
I have added a [PR](https://github.com/djangopackages/djangopackages/pull/1005) that should close this, but I think we could start a discussion about migrating to celery to run periodic tasks and some other tasks that can require significant amount of time such as `Fetch latest data` action.

jefftriplett:
> I have added a [PR](https://github.com/djangopackages/djangopackages/pull/1005) that should close this, but I think we could start a discussion about migrating to celery to run periodic tasks and some other tasks that can require significant amount of time such as `Fetch latest data` action.

+1

Moving the "fetch latest data" action and some of the crontabs one or a few at a time would be an immediate improvement IMO. 

Have you used `django-q2`? I'm using it on a few high traffic websites and it hasn't missed a beat. I have used Celery for >15 years so I'm comfortable with it, but I found q2 easier to work with and it requires a lot fewer resources to run. I could go either way, but I was tempted to use it over celery since the migration path isn't terrible.

saadmk11:
> > I have added a [PR](https://github.com/djangopackages/djangopackages/pull/1005) that should close this, but I think we could start a discussion about migrating to celery to run periodic tasks and some other tasks that can require significant amount of time such as `Fetch latest data` action.
> 
> +1
> 
> Moving the "fetch latest data" action and some of the crontabs one or a few at a time would be an immediate improvement IMO. 
> 
[...]

PR #1007, opened 2023-08-07
Migrate Long Running Tasks to `django-q`

saadmk11:
### TODO

- [x] Migrate long running Tasks to `django-q`
- [x] Add appropriate settings for `Q_CLUSTER`

**In different PR's:**

- Migrate Cron Jobs to `django-q`
- Remove Celery Integration and Cleanup Celery References

### Reference

Moving the "fetch latest data" action and some of the crontabs one or a few at a time would be an immediate improvement IMO. 

Have you used `django-q2`? I'm using it on a few high traffic websites and it hasn't missed a beat. I have used Celery for >15 years so I'm comfortable with it, but I found q2 easier to work with and it requires a lot fewer resources to run. I could go either way, but I was tempted to use it over celery since the migration path isn't terrible.

_Originally posted by @jefftriplett in https://github.com/djangopackages/djangopackages/issues/961#issuecomment-1666854165_


settings/base.py
```diff
@@ -355,3 +356,10 @@ def SOCIAL_AUTH_DEFAULT_USERNAME(u):
 # if True admin site will not be affected by the maintenance-mode page
 MAINTENANCE_MODE_IGNORE_ADMIN_SITE = True
 FORM_RENDERER = "django.forms.renderers.DjangoDivFormRenderer"
+
+# django-q2 settings
+# https://django-q2.readthedocs.io/en/stable/configure.html
+Q_CLUSTER = {
+    "name": "djangopackages",
+    "redis": env.str("REDIS_URL"),
+}
```
saadmk11, line 365: We need to set some configurations here depending on the production server (ideally using environment variables) such as `workers`

We also need to decide if we want to disable task retries (e.g: `ack_failures`). If we want to enable it then we should also set `timeout` and `retry` values.

C.C: @jefftriplett

jefftriplett: I think the default is 4 workers (or at least the example config is 4 workers) which seems right to me. 

I don't like how they say they sleep/block though. We might just have to start with defaults and adjust. Worse case we build setup two queues so that we always have one available that won't ever block. At the moment just processing tasks out of band of a web node is an improvement so I don't think we have to optimize for make it more complex that we need.

jefftriplett (approved):
This looks good to me. I know you have it marked as WIP, but I don't mind trying it out.

saadmk11:
> This looks good to me. I know you have it marked as WIP, but I don't mind trying it out. 

Sounds good 👍

saadmk11:
> I'm hoping to merge/deploy this tonight. Friday is my backup if tonight doesn't work out. 

Awesome 👍, let me know how it goes.

PR #1011, opened 2023-08-16
:fire: Removes Celery

jefftriplett:
This removes the last Celery bits, adds a 10-minute timeout, and makes a few Compose examples.


jefftriplett:
@saadmk11 going to do a post-merge review on this so I can silence a `timeout` that can't match `retry` warning that's spamming the console.

saadmk11:
> @saadmk11 going to do a post-merge review on this so I can silence a `timeout` that can't match `retry` warning that's spamming the console. 

That's a weird bug, even if you set timeout to None it will still match it with retry default even though the retry functionality will never be used. 🫤

jefftriplett:
Ya, you are right. It's not "fixed" but it's no longer flooding the logs. This is so isolated we can definitely re-think it when we start moving cronjobs over. That's why I went with the 10-minute timeout since we have a few jobs that can go for hours when it does a daily all-package rescan which is a good excuse to refactor it.

saadmk11:
> Ya, you are right. It's not "fixed" but it's no longer flooding the logs. This is so isolated we can definitely re-think it when we start moving cronjobs over. That's why I went with the 10-minute timeout since we have a few jobs that can go for hours when it does a daily all-package rescan which is a good excuse to refactor it.

There is a issue https://github.com/Koed00/django-q/issues/526 related to this bug on for this on the original repository which is unresolved.

I have added a PR https://github.com/djangopackages/djangopackages/pull/1012 to move our cron jobs to `django-q` which uses a separate timeout settings.

Issue #1026, opened 2023-09-27
grid pages broken


dogukanakin:
from web

<img width="1678" alt="Ekran Resmi 2023-09-27 14 54 28" src="https://github.com/djangopackages/djangopackages/assets/101465660/6d1fa259-eaff-4fca-b449-c2a6a75b0bd8">

from phone

![Django Packages](https://github.com/djangopackages/djangopackages/assets/101465660/bf5ad4a1-e429-4151-8742-d2450f9407f5)

the-ai-metiss:
Same issue here

jefftriplett:
Thank you for the report. I'm afraid my bus factor + four days of flu + my youngest having a minor procedure (they are fine) was the perfect storm for me not being around to see this.

PR #1027, opened 2023-09-27
:package: Adds crispy-bootstrap3 (hotfix)

### LINKED ISSUE #1021: all /grids/ views are broken
**Describe the bug**
Clicking on any of the grid listings at the top, for example [deployment](https://djangopackages.org/grids/g/deployment/) generates an error

**To Reproduce**
1. Go to Home Page
2. Click on `Deployment` (or any package listed at the top)
3. See error

**Expected behavior**
Clicking on deployment will list out the packages that are deployment related

**Screenshots**

image

jefftriplett:
Fixes #1021 #1026 #1025 

Our/my bus factor finally kicked in without Sentry messages making it to me. I was down with the flu for four days and then was mostly offline with my youngest having a minor procedure and came back to seeing an update broke something. 

We can re-assess and figure out how to make it easier to see these and fix them so it's not just deployed from my machine. 

Thanks for jumping in on these.

### DIFF
--- a/dockerfiles/redis/Dockerfile
+++ b/dockerfiles/redis/Dockerfile
@@ -1,3 +1,3 @@
-FROM redis:3.2
+FROM redis:7.2
 COPY redis.conf /usr/local/etc/redis/redis.conf
 CMD ["redis-server", "/usr/local/etc/redis/redis.conf" ]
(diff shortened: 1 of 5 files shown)


saadmk11:
We need to add `crispy_bootstrap3` to INSTALLED_APPS

jefftriplett:
@saadmk11 thanks. I noticed that too once I could run it locally. I'm on my work machine and everything is so slow trying from probably using non-Apple M2-friendly images. Gotta look at that next.

PR #1046, opened 2023-10-25
Move to Ruff Formatter

saadmk11:
Reference: https://docs.astral.sh/ruff/formatter/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ exclude = [
 per-file-ignores = {}
 
 # Same as Black.
-line-length = 120
+line-length = 88
 
 # Allow unused variables when underscore-prefixed.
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -57,7 +57,7 @@ exclude = [
 per-file-ignores = {}
 
 # Same as Black.
-line-length = 120
+line-length = 88
```
saadmk11, line 60: We were actually using `line-length = 88` (Black default)

jefftriplett:
For another project, I refactored to use https://github.com/westerveltco/django-email-relay/blob/main/pyproject.toml in case anything looks useful. I plan to compare/backport a few things for some other projects (their pre-commit config might be useful too)

PR #1047, opened 2023-10-28
Add back Grid of the Week

ryancheley:
This is the initial draft to implement #720 

This PR also adds back the Package of the Week, and the Public Service Announcement. If there are no Grids, Packages, or PSAs a message will be displayed indicating such


templates/homepage/django_package_of_the_week_partial.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}
+<div class="text-center categories col-sm-4">
+    <h2>{% trans "Featured Package" %}</h2>
+    {% if potw %}
+        <h4><a href="{{ potw.get_absolute_url }}">{{ potw.title }}</a></h4>
+        <p>{{ potw.description }}</p>
+        <p>
+            <strong>{% trans "Packages" %}:</strong> {{ potw.gridpackage_set.count }}
+            <strong>{% trans "Features" %}:</strong> {{ potw.feature_set.count }}
```
saadmk11, line 9: Package Model do not have direct relation to Feature Model.

templates/homepage/public_service_announcement_partial.html
```diff
@@ -0,0 +1,7 @@
+{% load i18n %}
+<div class="text-center categories col-sm-4">
+    <h2>{% trans "Public Service Announcement" %}</h2>
+    {% if psa_body %}
+        <p>{{ psa_body | safe }}</p>
+    {% endif %}
```
saadmk11, line 6: Maybe we can add some text if there are no Announcements?

templates/homepage/grid_of_the_week_partial.html
```diff
@@ -0,0 +1,14 @@
+{% load i18n %}
+<div class="text-center categories col-sm-4">
+    <h2>{% trans "Featured Grid" %}</h2>
+    {% if gotw %}
+        <h4><a href="{{ gotw.get_absolute_url }}">{{ gotw.title }}</a></h4>
+        <p>{{ gotw.description }}</p>
+        <p>
+            <strong>{% trans "Packages" %}:</strong> {{ gotw.gridpackage_set.count }}
+            <strong>{% trans "Features" %}:</strong> {{ gotw.feature_set.count }}
```
saadmk11, line 9: We can pre-calculate thees using aggregation.

homepage/views.py
```diff
@@ -234,25 +238,31 @@ def homepage(request, template_name="homepage.html"):
...
 
     # Public Service Announcement on homepage
-    # try:
-    #     psa_body = PSA.objects.latest().body_text
-    # except PSA.DoesNotExist:
-    #     psa_body = '<p>There are currently no announcements.  To request a PSA, tweet at <a href="http://twitter.com/open_comparison">@Open_Comparison</a>.</p>'
+    try:
+        psa_body = PSA.objects.latest().body_text
+    except PSA.DoesNotExist:
+        psa_body = '<p>There are currently no announcements.  To request a PSA, tweet at <a href="http://twitter.com/open_comparison">@Open_Comparison</a>.</p>'
```
saadmk11, line 265: Interesting, did not know something like this existed. Is `@Open_Comparison` twitter account active anymore?

saadmk11 (commented):
Awesome Work! :100:  Nice to see this feature being worked on again. :+1:

homepage/views.py
ryancheley: @saadmk11 I'm not sure. This was code that was already in the file that had been commented out. I put it back in with the intention of asking @jefftriplett what, if any, method of communication may be set up for this kind of thing. I'm not sure if there's an official DjangoPackages presence on either X or Mastodon. 

This may be something that can be tied into Django News potentially, but that's just an idea I was kind of kicking around.

templates/homepage/public_service_announcement_partial.html
ryancheley: There is text in the model that returns when there are no announcements, but it looks like this comment was added before this [one](https://github.com/djangopackages/djangopackages/pull/1047#discussion_r1375429334) so I think this can be resolbed

templates/homepage/grid_of_the_week_partial.html
ryancheley: @saadmk11 would you do that in the view or in the template? 

I can kind of see how I might start to get that idea set up in the view, but then my idea would kind of fall apart after that and get more complicated than it would probably need to be 😩

Any ideas on how to implement would be welcome 🙏

homepage/views.py
jefftriplett: I created https://fosstodon.org/@djangopackages and have used it sparingly to mainly communicate downtime. 

I'd be fine just not having a PSA if there isn't anything to shout out. That's probably the easiest way to manage it. 

(I'll check this out later tonight or the first of the week)

templates/homepage/grid_of_the_week_partial.html
saadmk11: We can add something like this
```python
potw = (
    Package.objects.filter(
        dpotw__start_date__lte=my_today, dpotw__end_date__gte=my_today
    )
    .annotate(grid_package_count=Count("gridpackage"))
    .only(
        "pk",
        "slug",
        "title",
        "repo_description",
    )
    .latest(
        "dpotw__end_date",
    )
)
```

ryancheley: @saadmk11 thanks! I was trying to overly complicate it but this will get me in the right spot 😁

homepage/views.py
jefftriplett: I carved out some time to check this out today and I quickly realized I need to develop a better plan for how to roll out the Grids and Packages. I'll work on this over the weekend and see which grids can be cleaned up to start targeting. I can do a follow-up about announcements too. 

@ryancheley thank you for getting this ink on the page aka championing these features to bring them back.

jefftriplett (approved):
This checks the right boxes. I'm going to follow up with a PR that added some switches around the new sections before we 🚀  deploy it. 

Good work!

PR #1048, opened 2023-10-29
Move Load Sample Data Instructions to Installation Page

saadmk11:
It made more sense to move Load Sample Data Instruction to Installation Page rather than keeping it on the FAQ page. This makes it easy to discover Load Sample Data Instructions

PR Doc Build: https://djangopackages--1048.org.readthedocs.build/en/1048/install/


jefftriplett (approved):
Excellent idea!

PR #1053, opened 2023-11-04
:sparkles: Feature flag grid and package of the week + psa

jefftriplett:
The PR adds a few feature switches around the Gride of the Week + Package of the Week. The PSF flag is primarily a place holder until we swap out some copy.

### DIFF
--- a/homepage/tests/test_views.py
+++ b/homepage/tests/test_views.py
@@ -43,7 +43,7 @@ from django.urls.exceptions import NoReverseMatch
 
 def test_homepage(db, tp, django_assert_num_queries):
     url = tp.reverse("home")
-    with django_assert_num_queries(10):
+    with django_assert_num_queries(15):
         response = tp.client.get(url)
     assert response.status_code == 200
 
--- a/templates/_homepage.html
+++ b/templates/_homepage.html
@@ -7,15 +7,15 @@
 
 
 {% switch "switch_use_grid_of_the_week" %}
-<div class="container">
-    <div class="row">
-        {% include "homepage/grid_of_the_week_partial.html" %}
-        {% include "homepage/django_package_of_the_week_partial.html" %}
-        {% switch "switch_use_psa" %}
-            {% include "homepage/public_service_announcement_partial.html" %}
-        {% endswitch %}
+    <div class="container">
+        <div class="row">
+            {% include "homepage/grid_of_the_week_partial.html" %}
+            {% include "homepage/django_package_of_the_week_partial.html" %}
+            {% switch "switch_use_psa" %}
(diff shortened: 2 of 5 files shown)


templates/_homepage.html
```diff
@@ -1,25 +1,35 @@
...
+        <div class="row">
+            {% include "homepage/grid_of_the_week_partial.html" %}
+            {% include "homepage/django_package_of_the_week_partial.html" %}
+            {% switch "switch_use_psa" %}
+                {% include "homepage/public_service_announcement_partial.html" %}
+            {% endswitch %}
+        </div>
     </div>
-</div>
+{% endswitch %}
```
ryancheley, line 19: I really like this. I haven't had the opportunity to use `waffle` before, but this makes me want to try it out 😄

homepage/tests/test_views.py
```diff
@@ -43,7 +43,7 @@
 
 def test_homepage(db, tp, django_assert_num_queries):
     url = tp.reverse("home")
-    with django_assert_num_queries(10):
+    with django_assert_num_queries(15):
```
ryancheley, line 46: @jefftriplett is the increase in expected number of queries because of the introduction of the use of waffle?

jefftriplett: I believe so. There's an option to use redis (I haven't read it in a while) but that's kind of a moot point when testing it. We can test with the flags off and on to narrow it down, but I was mostly looking for a way to ship your feature before dependabot starts burying us in updates :)

saadmk11 (approved):
Looks good 👍 we could add some documentation on where and how we use it if this is not a temporary thing.

templates/_homepage.html
```diff
@@ -1,25 +1,35 @@
...
-        {% include "homepage/django_package_of_the_week_partial.html" %}
-        {% include "homepage/grid_of_the_week_partial.html" %}
-        {% include "homepage/public_service_announcement_partial.html" %}
+
+{% switch "switch_use_grid_of_the_week" %}
+    <div class="container">
+        <div class="row">
+            {% include "homepage/grid_of_the_week_partial.html" %}
+            {% include "homepage/django_package_of_the_week_partial.html" %}
+            {% switch "switch_use_psa" %}
```
saadmk11, line 14: Does this mean we need both this and "switch_use_grid_of_the_week" switch on to make PSA visible?

jefftriplett: yes, exactly. We'll probably launch something for it before the PSAs. I'll update that text block to point to our mastodon since that's low-hanging fruit.

jefftriplett: Btw, these auto-register in the admin so there isn't a guessing game to find them.

PR #1073, opened 2023-12-02
📝Initial idea for docs generation

ryancheley:
There's still a bit that needs to be done, but as an initial proof of concept this seems like a good start. One thing I keep running into is that if I try to move the file `docs-tamplate.md` into the `docs` directory (where I think it **should** go) I get an error like this:

```
ModuleNotFoundError: No module named 'settings'
```

I tried various things to get rid of this error (like using `settings.configure()`) but this just threw error messages about missing settings (like LOGGING_CONFIG). Each time I would try and add a new setting I would get a new error about another missing setting. Not sure what is the best way to approach something like this.


saadmk11:
As we are using Mkdocs for documentation, we can probably checkout [mkdocstrings](https://mkdocstrings.github.io/).

This mkdocs plugin uses docs strings to generate documentation. (Though I am assuming this is what we are going for here without full context, so, I might be wrong :))

ryancheley:
@saadmk11 [mkdocstrings](https://mkdocstrings.github.io/) looks really interesting and seems to do what I was hoping to do. I’ll dig into it a bit more tomorrow. Thanks for pointing it out!

docker-compose.yml
```diff
@@ -28,6 +28,8 @@ services:
   postgres:
     # image: "pgautoupgrade/pgautoupgrade:latest"
     build: ./dockerfiles/postgres
+    env_file:
+      - .env.local
```
saadmk11, line 32: This fix has already been merged, A git rabase should fix the issue. :)

ryancheley:
> > I pushed up a change to use `mkdocstrings` (thanks @saadmk11 for the idea!) and I think it will work much better than the previous attempt. There's a bit of cleanup that is still needed, but am hoping to get some feedback on the attempt so far
> 
> This looks like a great start 👍 , I think we can move all the deprecation rule docs to a single file.

@saadmk11 I'm not sure what you mean by moving all of the deprecation rule docs to a single file. I believe they already are. What happened to me was I initially read `Description` as `Deprecated` and I thought they were duplicative. Not sure if that's what you're referring to as well 😄

saadmk11:
> > > I pushed up a change to use `mkdocstrings` (thanks @saadmk11 for the idea!) and I think it will work much better than the previous attempt. There's a bit of cleanup that is still needed, but am hoping to get some feedback on the attempt so far
> > 
> > 
> > This looks like a great start 👍 , I think we can move all the deprecation rule docs to a single file.
> 
> @saadmk11 I'm not sure what you mean by moving all of the deprecation rule docs to a single file. I believe they already are. What happened to me was I initially read `Description` as `Deprecated` and I thought they were duplicative. Not sure if that's what you're referring to as well 😄

I meant we can move all the Search Score Rules into the same docs page as they don't contain much text.
Screenshot from 2023-12-05 13-25-58

ryancheley:
@saadmk11 I think I've made the change you're suggesting. can you take a look?

jefftriplett:
This is looking good to me so far. We can always break rules into their own pages once we have the content for it. So this feels like we are landing in a good place to reference. 🎉

saadmk11:
> @saadmk11 I think I've made the change you're suggesting. can you take a look?

Thanks for the quick updates. Will do a full review after the PR is ready for review :)

jefftriplett (approved):
LGTM

PR #1005, opened 2023-08-05
Add Utility Service to Run CronJobs

saadmk11:
With this we should be able to run:
```shell
12 23 * * * /usr/bin/docker compose -f /code/djangopackages/docker-compose.prod.yml --profile utility run --rm utility python manage.py searchv2_build
36 17 * * * /usr/bin/docker compose -f /code/djangopackages/docker-compose.prod.yml --profile utility run --rm utility python manage.py package_updater
48 21 * * * /usr/bin/docker compose -f /code/djangopackages/docker-compose.prod.yml --profile utility run --rm utility python manage.py pypi_updater
```

closes: https://github.com/djangopackages/djangopackages/issues/961

PR #1075, opened 2023-12-09
📝 Added docs for django-q and utility

ryancheley:
django-q was introduced in #1007
utility was introduced in #1005


ryancheley:
@jefftriplett I was going through some of the docs trying to get more familiar with the stack and found this [page](https://docs.djangopackages.org/en/latest/deployment/#stack) but it was missing references to `django-q` and `utility` hence this PR. The docs also indicate that `ubuntu 14.04` is used. I recall in #873 there was a question about if the server was still on 14.04. If it isn't we should update the docs to the version that we're on.

jefftriplett:
@ryancheley I misread this as an issue from my phone. Happy to merge it and then update #1076 which expands/updates a bit deeper.

Issue #1014, opened 2023-08-21
Document production cron


jefftriplett:
> We can adjust cron in a new PR.

Do you mean the cron schedule?

---

How do we setup cron jobs on the current production servers?  I don't see any references to it on the codebase other than this:
https://github.com/djangopackages/djangopackages/blob/024ece9980e47868311f840a7daeee284925bcf2/fabfile.py#L127-L131

Do we set this up manually on production server by modifying the `crontab` file?

_Originally posted by @saadmk11 in https://github.com/djangopackages/djangopackages/issues/1012#issuecomment-1685891031_

PR #1076, opened 2023-12-10
:pencil: Updates deployment docs

jefftriplett:
Fixes #1075 

@ryancheley let me know what else you'd like to see here. Also, feel free to contribute in this branch if you see something you really want to change.


docs/docs/deployment.md
```diff
@@ -47,28 +38,6 @@ or via `just`:
 just deploy
 ```
 
-## Backups
```
jefftriplett, line 50: These are managed through Digital Ocean as part of their managed service. We do need to learn up our `justfile` to create a better pg_dump + pg_restore recipe, but production is isolated by design via DO.

docs/docs/deployment.md
```diff
@@ -2,35 +2,26 @@
...
-The configuration in `docker-compose.yml` contains 4 services:
+All of our `docker-compose*.yml` configurations contains the following services:
 
-> - `postgres` that powers the database
-> - `django-a` and `django-b` that runs the WSGI server and serves the app through gunicorn
-> - `caddy` that proxies incoming requests to the gunicorn server
-> - `redis` as cache
+- `postgres` powers our database.
+- `django` powers our Python and Django backend. In production we use call these `django-a` and `django-b` to run our WSGI server and serves the app through gunicorn.
+- `django-q` powers our task queues and background workers.
```
saadmk11, line 20: Did we start to use django-q in production which replaces the cron jobs? 

Related:
 - https://github.com/djangopackages/djangopackages/issues/1014

Can probably be closed after this is merged?

docs/docs/deployment.md
```diff
@@ -2,35 +2,26 @@
...
 ## Stack
 
-The configuration in `docker-compose.yml` contains 4 services:
+All of our `docker-compose*.yml` configurations contains the following services:
 
-> - `postgres` that powers the database
-> - `django-a` and `django-b` that runs the WSGI server and serves the app through gunicorn
-> - `caddy` that proxies incoming requests to the gunicorn server
-> - `redis` as cache
+- `postgres` powers our database.
```
saadmk11, line 18: Maybe we can mention that we are using managed PostgreSQL in production?

saadmk11 (approved):
Nice improvements 💯

docs/docs/deployment.md
jefftriplett: We are using both django-q and cron. 

After months of django-q running stable, I think we can keep moving jobs over to it until we are cron-less.

docs/docs/deployment.md
jefftriplett: Great idea. 

Side-note: I'd like to remove the `postgres/Dockerfile` at some too since that feels like part of a different era and I don't see where that's useful anymore.

saadmk11: Sounds good, created an issue to track this https://github.com/djangopackages/djangopackages/issues/1077

docs/docs/deployment.md
saadmk11: Would be happy to help with this if you need any 😊

docs/docs/deployment.md
```diff
@@ -2,35 +2,26 @@
...
+djangopackages.org use Docker Compose both for local development and in production.
 
 The current strategy is:
 
-> - Use a virtual machine with a well patched OS (debian, ubuntu, RHEL), djangopackages.org is using
->   ubuntu 14.04
-> - Install docker, docker-compose, git and supervisord
-> - Clone the code on the server
-> - Let supervisord run it
+- Use a virtual machine with a well patched OS (debian or ubuntu), djangopackages.org is using
```
ryancheley, line 9: This just kind of trails off and doesn't indicate what djangopackages.org is using.

ryancheley (approved):
Other than the one question about the version of the OS being used. 👍

docs/docs/deployment.md
jefftriplett: We are more tied to an app platform than a host OS these days so I'm not sure we have a set host requirement. The only reason I haven't tried running it on DO's app platform is that it would be more expensive. Open to ideas though.

PR #1080, opened 2023-12-15
Use `pgautoupgrade` Docker Image for `postgres` Docker Service

### LINKED ISSUE #1077: Use Official PostgreSQL Docker Image for "postgres" Docker Service
I'd like to remove the `postgres/Dockerfile` at some too since that feels like part of a different era and I don't see where that's useful anymore. 

_Originally posted by @jefftriplett in https://github.com/djangopackages/djangopackages/pull/1076#discussion_r1421759886_

saadmk11:
closes #1077


dockerfiles/postgres/backup.sh
```diff
@@ -1,30 +0,0 @@
-#!/bin/bash
```
saadmk11, line 1: I removed this with the assumption that we don't use these anymore. I can revert the changes if these are in-use (Though we probably need to use a custom Docker image then).

C.C: @jefftriplett

docker-compose.yml
```diff
@@ -26,30 +28,30 @@ services:
     init: true
 
   postgres:
-    # image: "pgautoupgrade/pgautoupgrade:latest"
-    build: ./dockerfiles/postgres
+    image: "pgautoupgrade/pgautoupgrade:latest"
```
saadmk11, line 31: Could not find `postgres-13.x` (assuming we are using `postgres-13.x` in production) version of `pgautoupgrade`. I would have preferred to use the same version as production for local development.

fabfile.py
```diff
@@ -100,14 +100,6 @@ def rollback(commit="HEAD~1"):
     deploy()
 
 
-def backup():
```
saadmk11, line 103: Same reason: https://github.com/djangopackages/djangopackages/pull/1080/files#r1428324217

jefftriplett (approved):
@saadmk11 looks good to me.

Issue #1086, opened 2023-12-22
:fire: Pull Carbon Ads


jefftriplett:
I'll update the payout stats, but we went from almost breaking even on hosting costs ~$70 a month down to ~$20 a month which wasn't worth while.

jefftriplett:
Fixed in #1087

PR #1087, opened 2023-12-22
:sparkles: Friday updates

jefftriplett:
- :pencil: Updates costs/income for 2023
- :fire: Removes Carbon Ads #1086


jefftriplett:
Swapping out Ads and updating.

PR #1110, opened 2024-02-23
:sparkles: Add Blog app + model

jefftriplett:
My goal is to work up to a simple `/changelogs/` blog-like thing.


blog/models.py
```diff
@@ -0,0 +1,33 @@
...
+class Post(BaseModel):
+    title = models.CharField(max_length=200)
+    slug = models.SlugField(
+        blank=True,
+        max_length=250,
+        null=True,
+        unique_for_date="published_date",
+    )
+    content = models.TextField()
+    author = models.ForeignKey(User, blank=True, on_delete=models.CASCADE, null=True)
```
saadmk11, line 17: Do we want to delete the blog post if the author is deleted?

saadmk11 (commented):
Changes look good!

blog/models.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Post(BaseModel):
+    title = models.CharField(max_length=200)
+    slug = models.SlugField(
+        blank=True,
+        max_length=250,
+        null=True,
+        unique_for_date="published_date",
+    )
+    content = models.TextField()
```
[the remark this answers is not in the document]
jefftriplett: Do you have one you like? I have been through a dozen or so, and I end up writing things in Obsidian and copy + pasting them in for decent markdown support.

blog/admin.py
```diff
@@ -0,0 +1,9 @@
+from django.contrib import admin
+
+from blog.models import Post
+
+
+@admin.register(Post)
+class PostAdmin(admin.ModelAdmin):
+    list_display = ["title", "published_date"]
+    raw_id_fields = ["author"]
```
ryancheley, line 9: I had no idea about `raw_id_fields` and now there is a project I need to go revisit so I can implement this!

blog/models.py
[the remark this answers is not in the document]
ryancheley: @jefftriplett is the idea that all blog posts would be written from the admin?

blog/models.py
ryancheley: Are users ever hard deleted, or are they deactivated? If they aren't ever deleted then I'm not sure how much this would matter, but if they are deleted then there would be a risk that articles would be deleted when the user was deleted (but I'm pretty sure the admin would give a warning about the associated records, right?)

blog/models.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def __str__(self):
+        return self.title
+
+    def save(self, *args, **kwargs):
+        # If the post is being published for the first time, generate a slug
+        if self.published_date and not self.slug:
+            # Format the published_date to include only the date portion
+            date_str = self.published_date.strftime("%Y-%m-%d")
+            self.slug = slugify(f"{date_str}-{self.title}")
```
ryancheley, line 31: Would we want to have the date in the slug, or in the path of the url? Something like this

`2024/02/24/another-test`

vs this

`2024-02-24-another-test`

I'm not sure if there is an SEO hit (or if we care if there is)

ryancheley:
I like this change a lot. Great work!

ryancheley (approved):

blog/models.py
jefftriplett: Good question. I have been eyeballing https://obsidian.md/changelog/ and leaning towards `2024-02-24-another-test` for inspiration. I have a List + Detail view on the backend to keep things easier.

blog/models.py
jefftriplett: I changed this to an optional char field. I don't have it exposed in the templates, but this at least addresses one concern.

PR #1117, opened 2024-03-14
:tractor: Switch to uv --system

jefftriplett:
This updated `uv pip install` to skip the `venv` step, updates a few warnings, and makes a few `justfile` recipes along the way. 

Some context: https://micro.webology.dev/2024/03/14/pythons-uv-tool.html


justfile
```diff
@@ -213,39 +220,32 @@ bootstrap *ARGS:
...
     just pre-commit --all-files
 
-# TODO: Make the target-version a variable
-
-# Run `django-upgrade` with a target version of 4.1
+# Upgrade existing Python dependencies to their latest versions
 @upgrade:
-    git ls-files -- "*.py" | xARGS django-upgrade --target-version=4.1
+    just lock --upgrade
+    # git ls-files -- "*.py" | xARGS django-upgrade --target-version=4.1
```
saadmk11, line 248: Target version 5.0?

saadmk11 (approved):
Nice 👍

saadmk11:
I think uv doesn't need pip to be installed either. We can directly install uv instead at some point.

justfile
jefftriplett: Yes, but I think this was carry over from before pre-commit. I'll clean it up in a follow up PR since it should run when we lint.

jefftriplett:
> I think uv doesn't need pip to be installed either. We can directly install uv instead at some point.

Great point. I have struggled not to upgrade the pip while I'm installing uv. In theory, I think you are right, and pip is really only used to get `uv` installed.

Issue #1115, opened 2024-03-09
✨ Update Profile page options


jefftriplett:
I noticed it's been a while since we updated our profile page. 

Both the detail page: https://djangopackages.org/profiles/jeff/
And then edit my profile page: https://djangopackages.org/profiles/edit/

- [x] Add GitLab as an option
- [x] Add Bitbucket as an option
- [x] We should normalize the spelling of GitHub (vs. Github)
- [x] Remove the Google Code `google_code_url`
- [x] Fix the `raw_id_fields` on the Profile model admin to avoid all ~15k users being rendered in a drop-down
- [x] Add social media or custom links? 
- [x] Create a changelog post about the new feature. 

We have room to expand our profile options if we want to add a description or any social websites.  

Also, see this thread for a potential follow-up feature to let people list their favorite packages: https://mastodon.social/@webology/112066712825122434

ryancheley:
@jefftriplett for the favorite packages is the idea that IF a user has starred a project that is listed in Django Packages it would appear on their profile page, or would they have to actually add the favorite package from the list of one available from Django Packages? 

For the Social Media or Custom links, we could take a page from Mastodon and have an `extra fields` section.

<img width="556" alt="image" src="https://github.com/djangopackages/djangopackages/assets/9857779/7d560af3-3c90-4a10-b3e7-5a67d153521d">

jefftriplett:
@ryancheley I like that a lot. I'm not sure if we are asking for more spam, but we can mitigate some of that by using `rel="nofollow"`.

ryancheley:
@jefftriplett I have some ideas on how to implement this and will take a swing at making it work 😄

jefftriplett:
@ryancheley yes plz. Ping me if you need a body to throw ideas at. I have no real implementation plan in mind outside of, Mastodon is a good place to get inspiration from.

jefftriplett:
@ryancheley tests run clean to me. I run `pytest --create-db` when there are schema changes. 

How do you feel about making them their own table and adding them as an inline in the form? Then we could set it maybe show max of 4 (with room to expand if we wanted)?

I was curious how Mastodon does, and as best as I could tell, they use a JSON field. https://github.com/mastodon/mastodon/blob/main/db/schema.rb#L186 

Not sure I'd want to do that but it's interesting.

ryancheley:
[...]

PR #1119, opened 2024-03-15
Update User Profile

saadmk11:
This PR only adds the action items presented in 
- https://github.com/djangopackages/djangopackages/issues/1115


ryancheley (approved):
looks good!

jefftriplett:
👀  This looks great. I'll give it a better pass this afternoon and then merge + deploy it. 

Would there be any fallout from changing the shape of the v3 API response? I don't want to roll a v4 to add/remove a field, but that's the only thing that popped up when I was reviewing.

PR #1143, opened 2024-04-19
:construction: :sparkles: Adds support for django-tailwind-cli

jefftriplett:
This needs to download and build tailwind in the production image before it's even remotely close to shipping.


ryancheley (approved):
@jefftriplett this look 💯 . Can't wait to migrate over to Tailwind!

saadmk11 (approved):
Looks good 🚀

jefftriplett:
It's closer but there are several places where Bootstrap and Tailwind aren't quite playing well together. See the form and the category styles. 

Xnapper-2024-04-26-16 19 26-96E72340-8763-4C1E-BBBE-C710445F5060

requirements.txt
```diff
@@ -1351,13 +1293,10 @@ sniffio==1.3.1 \
 social-auth-app-django==5.4.1 \
```
ryancheley, line 1293: I think with the pining of social-auth-app-django==5.4.0 this won’t be an issue, but wanted to call it out just in case

jefftriplett:
Doing some client work then hoping to resume this at some point this weekend.

Issue #1146, opened 2024-04-27
🚜 Convert sitemaps from a Django templates 


jefftriplett:
I noticed tonight that our sitemap.xml is a home `rolled` view + template. We should port this over to `django.contrib.sitemaps` and ideally use a [sitemap index](https://docs.djangoproject.com/en/5.0/ref/contrib/sitemaps/#creating-a-sitemap-index) to help cut down on the overall rendering time. 

- [ ] Static pages
- [ ] Packages 
- [ ] Grids
- [ ] Blog

See template: https://github.com/djangopackages/djangopackages/blob/main/templates/sitemap.xml#L1-L24

jefftriplett:
@saadmk11 great work. Here's it working in production: https://djangopackages.org/sitemap.xml

saadmk11:
> @saadmk11 great work. Here's it working in production: https://djangopackages.org/sitemap.xml 

Awesome 🍻

jefftriplett:
@saadmk11 I added it the changelog. 

https://djangopackages.org/changelog/2024-04-28-revamping-sitemap-generation-on-django-packages/

PR #1147, opened 2024-04-28
Move Sitemaps to Django Sitemap Framework

saadmk11:
closes https://github.com/djangopackages/djangopackages/issues/1146

**Note:** The Site Object (http://127.0.0.1:8000/admin/sites/site/) needs to be updated with proper domain name. i.e: `djangopackages.org`


ryancheley (approved):
Looks great! 💯

package/sitemaps.py
```diff
@@ -0,0 +1,13 @@
+from django.contrib.sitemaps import Sitemap
+from .models import Package
+
+
+class PackageSitemap(Sitemap):
+    changefreq = "daily"
+    priority = 1.0
```
jefftriplett, line 7:
```suggestion
    priority = 1.0
    protocol = "https"
```

grid/sitemaps.py
```diff
@@ -0,0 +1,13 @@
+from django.contrib.sitemaps import Sitemap
+from .models import Grid
+
+
+class GridSitemap(Sitemap):
+    changefreq = "daily"
+    priority = 1.0
```
jefftriplett, line 7:
```suggestion
    priority = 1.0
    protocol = "https"
```

urls.py
```diff
@@ -24,6 +29,15 @@
 admin.site.site_header = admin_header
 admin.site.site_title = admin_header
 
+SITEMAPS_CACHE_TTL = 24 * 60 * 60  # 24 Hours
```
jefftriplett, line 32: Let's drop this to 12 hours and see what it looks like. I have been bitten by 24 caching + once-a-day check, meaning ~2-day lag. 

```suggestion
SITEMAPS_CACHE_TTL = 12 * 60 * 60  # 24 Hours
```

jefftriplett (changes requested):
This looks great! I had only one nitpick on the protocol but I left suggestions for everything else. I'm happy to merge and deploy this after those get merged in. Good work!

grid/sitemaps.py
ryancheley: Since we’re on Django 5 do we need to be explicit with the protocol? Per the [docs](https://docs.djangoproject.com/en/5.0/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.protocol)

> Changed in Django 5.0:
> In older versions, the default protocol for sitemaps built outside the context of a request was 'http'.

saadmk11: @jefftriplett I intensionally left it as is. Because:

> If protocol is None, the URLs in the sitemap will use the protocol with which the sitemap was requested.

This will make the views generate proper URLs both in production and in development. (Assuming we don't allow http on production)

Ref: https://github.com/django/django/blob/8c257cecffc314179a5375ab8a16ca0ff3b6fb16/django/contrib/sitemaps/__init__.py#L13

jefftriplett: Perfect. After ~18 years of it not working, I'm thrilled to learn that it finally "just works"

jefftriplett (approved):

PR #1158, opened 2024-05-21
:fire: Remove user api crawling

jefftriplett:
I'm leaving this api in but requiring auth after seeing too many API 500 errors. I'm not a fan of having an open api endpoint for things like users.


saadmk11 (approved):

saadmk11:
We should deprecate APIv3 in favor of APIv4.

Issue #1065, opened 2023-11-18
Add tasks to pull in pypi stats and any other missing data points


ryancheley:
**Is your feature request related to a problem? Please describe.**
The current search results cant be swamped by historic stars on repositories making it difficult to find current packages with confidence

**Describe the solution you'd like**
Add tasks to pull in pypi stats and any other missing data points (to be determined)

jefftriplett:
(this should probably be in a new ticket but since it's related to pypi, I am adding it here at least to start)

I heard someone mention an interesting data point about how many dependencies your package depends on as being some measure of potential threat risks. 

In theory, we could use https://pypi.org/pypi/django/json and the `requires_dist` key to count how many dependencies the average package has. 

```json
"requires_dist":["asgiref<4,>=3.7.0","sqlparse>=0.3.1","tzdata; sys_platform == \"win32\"","argon2-cffi>=19.1.0; extra == \"argon2\"","bcrypt; extra == \"bcrypt\""],
```

Side note: This would very quickly spiral into counting the `requires_dist` of every package's packages for the threat risk, but I think it could still be useful with one level. (maybe two levels down the road if most of the packages are already cached)

Issue #834, opened 2022-08-10
Integrate monthly pypi-data into stats


jefftriplett:
The `pypi-data` has some nice information that we can use to expand some of our metrics to make search and rankings better. https://github.com/sethmlarson/pypi-data

PR #1160, opened 2024-05-24
:sparkles: Adds pypi_update_top_package_stats management command

jefftriplett:
Related to #1065 and #834

This pulls data for the top 8000 Python packages which hits about 4 to 5% of our packages which isn't great. 

I have another management command that requires Seth's monthly file download, but it's a little clunky.

### DIFF
--- a/package/management/commands/pypi_update_top_package_stats.py
+++ b/package/management/commands/pypi_update_top_package_stats.py
@@ -1,0 +1,46 @@
+import djclick as click
+import httpx
+
+from package.models import Package
+from rich import print
+
+
+DOWNLOAD_PATH = "top-pypi-packages-30-days.min.json"
+DOWNLOAD_URL = "https://raw.githubusercontent.com/hugovk/top-pypi-packages/main/top-pypi-packages-30-days.min.json"
+
+
+def normalize_pypi_slug(url: str) -> str:
+    url = url.replace("http:", "https:")
+    url = url.strip("/")
+    url = url.split("/")[-1]
+    return url
+
+
+@click.command()
+@click.argument("url", type=str, default=DOWNLOAD_URL)
+def command(url):
+    package_lookup = {}
+
+    request = httpx.get(url)
+    request.raise_for_status()
+
+    data = request.json()
+    print(f"last_update: {data['last_update']}")
+
+    rows = data["rows"]
+    for row in rows:
+        project = row["project"]
+        download_count = row["download_count"]
+        package_lookup[project] = download_count
+
+    objs = []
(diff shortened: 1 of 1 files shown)


package/management/commands/pypi_update_top_package_stats.py
```diff
@@ -0,0 +1,45 @@
...
+        project = row["project"]
+        download_count = row["download_count"]
+        package_lookup[project] = download_count
+
+    objs = []
+    packages = Package.objects.exclude(pypi_url__in=[None, ""]).order_by("pypi_url")
+    for package in packages:
+        pypi_slug = normalize_pypi_slug(package.pypi_url)
+        if pypi_slug in package_lookup:
+            package.pypi_downloads = package_lookup.get("pypi_slug", 0)
```
saadmk11, line 40: If the package already has non-zero value for `pypi_downloads` field then this will set the value to `0` if the project does not exist on the top list.

package/management/commands/pypi_update_top_package_stats.py
```diff
@@ -0,0 +1,45 @@
...
+
+    rows = data["rows"]
+    for row in rows:
+        project = row["project"]
+        download_count = row["download_count"]
+        package_lookup[project] = download_count
+
+    objs = []
+    packages = Package.objects.exclude(pypi_url__in=[None, ""]).order_by("pypi_url")
+    for package in packages:
```
saadmk11, line 37:
```suggestion
    for package in packages.iterator():
```

package/management/commands/pypi_update_download_package_stats.py
```diff
@@ -0,0 +1,95 @@
...
+def normalize_pypi_slug(url: str) -> str:
+    url = url.replace("http:", "https:")
+    url = url.strip("/")
+    url = url.split("/")[-1]
+    return url
+
+
+@click.command()
+@click.argument("url", type=str)
+@click.argument("filename", type=str)
```
saadmk11, line 50: Do we need this argument?  Seems like we are getting the filename from the `url` argument. i.e:

`filename = normalize_pypi_slug(url=url)`

saadmk11 (commented):
Only 4 to 5% hits is quite a small amount. :( It would have been great if there was a PyPI data collection for Django ecosystem specifically.

jefftriplett:
> Only 4 to 5% hits is quite a small amount. :( It would have been great if there was a PyPI data collection for Django ecosystem specifically.

Agreed. That's why I added the other tool but it's kind of annoying if everything isn't automated. Definitely doable with a v2.

Issue #1179, opened 2024-06-22
Pull the documentation link from PyPI's metadata if Package.documentation_url is None and the value exists.


FarhanAliRaza:
**Describe the solution you'd like**
Extract documentation url from pypi json api.

PR #1180, opened 2024-06-22
Extract documentaion_url from pypi

FarhanAliRaza:
Fixes https://github.com/djangopackages/djangopackages/issues/1179

### DIFF
--- a/package/models.py
+++ b/package/models.py
@@ -397,12 +397,12 @@ class Package(BaseModel):
                     elif docs_url := info.get("docs_url"):
                         self.documentation_url = docs_url
                     elif (
-                        info.get("desctiption")
+                        info.get("description")
                         and info.get("description_content_type") == "text/markdown"
                     ):
                         self.documentation_url = (
                             extract_documentation_url_from_markdown(
-                                info.get("desctiption")
+                                info.get("description")
                             )
                         )
 
--- a/package/models.py
+++ b/package/models.py
@@ -28,7 +28,12 @@ from core.utils import STATUS_CHOICES, status_choices_switch
 from package.managers import PackageManager
 from package.repos import get_repo_for_repo_url
 from package.signals import signal_fetch_latest_metadata
-from package.utils import get_pypi_version, get_version, normalize_license
+from package.utils import (
+    get_pypi_version,
+    get_version,
+    normalize_license,
+    extract_documentation_url_from_markdown,
+)
 
 repo_url_help_text = settings.PACKAGINATOR_HELP_TEXT["REPO_URL"]
 pypi_url_help_text = settings.PACKAGINATOR_HELP_TEXT["PYPI_URL"]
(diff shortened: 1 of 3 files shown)


package/models.py
```diff
@@ -375,6 +380,24 @@ def fetch_pypi_data(self):
                 if self.pypi_downloads is None:
                     self.pypi_downloads = total_downloads
 
+                # get documents_url from pypi
+                if self.documentation_url is None:
+                    if info["project_urls"].get("Documentation"):
```
saadmk11, line 385: After looking at the PyPI API, it seems like the name of this field (`"Documentation"`) can be anything. So, I think we can try all the common variants as well. e.g: `docs`

package/utils.py
```diff
@@ -72,3 +72,13 @@ def normalize_license(license: str):
             return "Custom"
         return stripped_license
     return "UNKNOWN"
+
+
+def extract_documentation_url_from_markdown(description):
```
saadmk11, line 77: Calling this function with `description="[Documentation](This is not a URL)"` will return an invalid URL. If we **really** need to extract the documentation URL from the `description` perhaps we could use some kind of `regex`? 

**Note:** A really large description filed might cause some performance issues.

package/models.py
```diff
@@ -375,6 +380,24 @@ def fetch_pypi_data(self):
...
+                        self.documentation_url = info["project_urls"].get(
+                            "Documentation"
+                        )
+                    elif info.get("docs_url"):
```
saadmk11, line 389: Could we check this field first?

saadmk11 (commented):
Hi @FarhanAliRaza, excellent work on this. :100: 

I just added some small nitpicks and suggestions. :)

package/models.py
FarhanAliRaza: I have seen that it is null for all the packages, I have seen but the name suggests it should contain the docs URL. I could not find any documentation for fields of JSON API. Can you share if you have any?. We can remove it if you think.

package/utils.py
FarhanAliRaza: Django has its own `URLValidator` Can we use that? URL regex gets pretty complex.

FarhanAliRaza: I did a simple test. How can I improve it?
image

package/models.py
```diff
@@ -375,6 +380,32 @@ def fetch_pypi_data(self):
...
+                        self.documentation_url = docs_url
+                    elif (
+                        info.get("description")
+                        and info.get("description_content_type") == "text/markdown"
+                    ):
+                        self.documentation_url = (
+                            extract_documentation_url_from_markdown(
+                                info.get("description")
+                            )
+                        )
```
jefftriplett, line 407: I think we can limit the scope to only what is structured data from PyPI JSON API. 

We have another tool we worked on which maintainers can use to help them score better that we still need to hook up.

package/models.py
saadmk11: Interesting, Let's keep it for now.

package/models.py
saadmk11: I agree.

FarhanAliRaza: removed

jefftriplett (approved):

jefftriplett:
@FarhanAliRaza good work! I tested it locally and it was hitting the correct values. I'll push a test, and then going forward, we can go over tests. 

The only thing that's missing would be the nightly pypi updater knowing to hit the right endpoints. I never speced that out fully, but I can.

Issue #1201, opened 2024-07-17
:elephant: Favorite Packages: Favorite Model


jefftriplett:
Please see #1170 for an overview. 

- Please create `favorites` app. 

- This model should connect a `User` to the `Package` model and should inherit from `core.models.BaseModel`. See here for details: https://github.com/djangopackages/djangopackages/blob/main/core/models.py#L7

- The `Favorite` model should be registered in `admin.py` and include `raw_id_fields` for both the `User` and the `Package` models.

## These two features depend on changes from #1203, but they can be a follow-up pull request, depending on the timing. 

- Override the `save` method to increment the `favorite_count` of the associated `Package` when a new favorite is created.

- Override the `delete` method to decrement the `favorite_count` when a favorite is removed.

PR #1204, opened 2024-07-17
Favourite model and facourites app

FarhanAliRaza:
fixes https://github.com/djangopackages/djangopackages/issues/1201


settings/base.py
```diff
@@ -130,6 +130,7 @@
     "blog",
     "classifiers",
     "core",
+    "favourites",
```
jefftriplett, line 133: Please change this and the rest of the app name from the British spelling of favourites to the American spelling of favorites. I'm happy to assist in the rename if that causes your database any issues.

jefftriplett (changes requested):
Lets please rename the app from the British English spelling to the American English spelling,

jefftriplett (approved):
Great work. Thank you for making the change.

favorites/models.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Favorite(BaseModel):
+    favorited_by = models.ForeignKey(
+        User, blank=True, null=True, on_delete=models.SET_NULL
+    )
+    package = models.ForeignKey(Package, on_delete=models.CASCADE)
+
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
+        self.package.favorite_count += 1
```
jefftriplett, line 15: Not sure if you have worked with `F()` expressions before. https://docs.djangoproject.com/en/5.0/ref/models/expressions/#f-expressions

They are really nice for changing relative values because Django will create an SQL statement that modifies the value that is stored in the database table at the time the expression is run versus whatever values are stored in the package.favorite_count at the time of the change.

PR #1225, opened 2024-07-26
:tractor: Re-stage TailwindCSS

jefftriplett:
Replaces #1143 

This seems to be working again. We need to run tailwind build with Docker, but I'm going to ship it and we can how it does in production.


jefftriplett:
@FarhanAliRaza This lands TailwindCSS integration with the project. Definitely room here to try it out and see where we could use it more as we move off of bootstrap.

PR #1226, opened 2024-07-26
:bug: Fixes a last_updated is None issue

jefftriplett:
Fixed a bug when running `python -m manage searchv3_build` with only `load_dev_data` generated data in the database.


jefftriplett:
cc @lipemorais on this. This was related to getting some of the search v3 ducks in a row.

PR #1230, opened 2024-07-31
Favorite Packages: Modify User Profile view

### LINKED ISSUE #1214: :sparkles: Favorite Packages: Modify User Profile view
Update the existing user profile view to include favorite packages.

## Details: 

  - This modification should:
     - Retrieve the user's favorite packages if `share_favorites` is True.
     - Add the list of favorite packages to the context data.
  - Update the user profile template to display the favorite packages.

## Additional Considerations (may apply to all views):

- Implement proper authentication check in all views.
- Add appropriate error handling and user feedback (e.g., messages framework).
- Update the package detail template to include a form for favoriting/unfavoriting.
- Ensure all views are tested thoroughly.
- Consider adding a confirmation step before unfavoriting a package.
- Consider using htmx to favorite/remove a favorite. 

---- 

See #1170 for more details.

vcruces:
- Modify user profile view to:
  - Retrieve the user's favorite packages if share_favorites is True.
  - Add the list of favorite packages to the context data.
- Update the user profile template to display the favorite packages.

Fixes #1214 

Screenshot 2024-07-30 at 21 47 22

(In draft because I have problems running unit tests in my local environment)

### DIFF
--- a/profiles/tests/test_views.py
+++ b/profiles/tests/test_views.py
@@ -66,6 +66,8 @@ class TestProfile(TestCase):
         self.assertEqual(p.gitlab_url, "zerg")
 
     def test_view_with_favorite_packages(self):
+        self.profile.share_favorites = True
+        self.profile.save()
         category = Category.objects.create(
             title="Test Favorite",
             slug="test_favorite",
--- a/profiles/tests/test_views.py
+++ b/profiles/tests/test_views.py
@@ -3,6 +3,8 @@ from django.test import TestCase
 from django.urls import reverse
 
 from core.tests.data import STOCK_PASSWORD, create_users
+from favorites.models import Favorite
+from package.models import Package
 from profiles.models import Profile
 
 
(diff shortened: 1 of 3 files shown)


profiles/tests/test_views.py
```diff
@@ -62,3 +64,24 @@ def test_edit_ryan(self):
...
+        Favorite.objects.create(package=package, favorited_by=self.user)
+        self.assertTrue(
+            self.client.login(username=self.user.username, password=STOCK_PASSWORD)
+        )
+        url = reverse(
+            "profile_detail", kwargs={"github_account": self.profile.github_account}
+        )
+        response = self.client.get(url)
+        self.assertContains(response, "Profile for user")
+        self.assertContains(response, "Favorite packages")
```
saadmk11, line 86: This assertion is failing because the profile probably does not have `share_favorites` set to `True`. We should set `self.profile.share_favorites=True` and save the profile before sending the request.

saadmk11 (commented):
Nice Work @vcruces :rocket: 

Regarding the user profile should we limit the number of packages displayed on each of the list? Because we don't have pagination here and there could be users with hundreds of favorite packages. C.C @jefftriplett

profiles/views.py
```diff
@@ -26,11 +27,17 @@ def profile_detail(request, github_account, template_name="profiles/profile.html
...
-    )
+    context = {
+        "local_profile": profile,
+        "user": profile.user,
+        "extra_fields": extra_fields,
+    }
+    if profile.share_favorites:
+        context["favorite_packages"] = Package.objects.filter(
+            favorite__favorited_by=profile.user
+        )
```
jefftriplett, line 38: Let's limit this to the top 25 packages. We can add pagination later. 

```suggestion
        context["favorite_packages"] = Package.objects.filter(
            favorite__favorited_by=profile.user
        )[:25]
```

jefftriplett (changes requested):
Looks good on a first pass. I see one change and one suggestion to add, and then we are ready to ship it! 

Nicely done!

profiles/tests/test_views.py
vcruces: You're right, that's the reason for the error, thanks!

profiles/views.py
vcruces: I'll do that! Thanks Jeff!

saadmk11 (approved):
🚀

jefftriplett (approved):

jefftriplett:
@vcruces great job!

Issue #1227, opened 2024-07-26
🔘  Create a favorite package feature flag


jefftriplett:
Let's add a [Waffle Feature flag](https://waffle.readthedocs.io/en/stable/types/flag.html) around the feature flag on the package details page so we can turn them on privately for testing. 

Django Waffle is already installed so we just need to configure it on the `templates/package/package.html`.

FarhanAliRaza:
have we created a flag for this feature ? What is the name of the flag?

PR #1233, opened 2024-07-31
favorite package feature flag

FarhanAliRaza:
fixes https://github.com/djangopackages/djangopackages/issues/1227

### DIFF
--- a/package/tests/test_views.py
+++ b/package/tests/test_views.py
@@ -102,7 +102,7 @@ class FunctionalPackageTest(TestCase):
 
     def test_package_detail_view(self):
         url = reverse("package", kwargs={"slug": "testability"})
-        with self.assertNumQueries(16):
+        with self.assertNumQueries(21):  # 16 expected 21 because of WAFFLE FLAG
             response = self.client.get(url)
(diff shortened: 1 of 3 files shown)


saadmk11 (commented):
A test is failing probably because `django-waffle` `templatetag` performs a database query which breaks the `assertNumQueries` assertion on that test.

Should we set `WAFFLE_CREATE_MISSING_FLAGS = True` in Django settings?

jefftriplett:
> Should we set `WAFFLE_CREATE_MISSING_FLAGS = True` in Django settings?

Yes plz.

package/tests/test_views.py
```diff
@@ -102,7 +102,7 @@ def test_package_list_view(self):
 
     def test_package_detail_view(self):
         url = reverse("package", kwargs={"slug": "testability"})
-        with self.assertNumQueries(16):
+        with self.assertNumQueries(21):  # 16 expected 21 because of WAFFLE FLAG
```
jefftriplett, line 105: This is a strangely high number. 🤔

jefftriplett (approved):
Looks good!

PR #1237, opened 2024-08-01
fix #1234: Add FavoritePackageRule to score favorite packages

### LINKED ISSUE #1234: Add FavoritePackageRule
Create a `FavoritePackageRule(ScoreRule)` rule, which we can use to score a package based on its number of favorites. 

Here are some example rules: 

- https://github.com/djangopackages/djangopackages/blob/main/searchv2/rules.py#L111C33-L132 
- https://github.com/djangopackages/djangopackages/blob/main/searchv2/rules.py#L111-L132

We can start with a `max_score: int = 20` to test it out. I suspect we will end up lowering this later. 

We can use this in our new search and it's runnable by adding it to this management command: https://github.com/djangopackages/djangopackages/blob/main/searchv2/management/commands/demo_rules_v3.py

lipemorais:
This new rule evaluates packages to determine if they are marked as favorite. It assigns a score based on the package's favorite status and provides relevant messages accordingly. This improvement helps refine the package scoring system.


searchv2/rules.py
```diff
@@ -132,6 +132,28 @@ def check(self, package: Package) -> CheckResult:
...
+    max_score: int = 20
+    documentation_url: str = f"{DOCS_URL}/rules/favorite-package"
+
+    def check(self, package: Package) -> CheckResult:
+        """
+        Check if the package is favorite.
+        Returns a full score and a success message if the package is not favorite,
+        or a zero score and an error message otherwise.
+        """
+        if not package.is_deprecated:
```
jefftriplett, line 151: We should check your favorites here instead of `is_deprecated`.

lipemorais: hmm, It's not clear to me how I should check the favorite here.


I'm thinking about creating an `is_favorite`, but I don't know what should be checked in this method/property.

jefftriplett: > hmm, It's not clear to me how I should check the favorite here.
> 
> I'm thinking about creating an `is_favorite`, but I don't know what should be checked in this method/property.

Good question. For now, `package.favorite_count > 0` is good enough. Once we get some real counts we can maybe create a sliding scale for the points, but this is fine for now. 

https://github.com/djangopackages/djangopackages/blob/main/package/models.py#L105

Note: I plan to adjust the `max_score` value before this goes into production.

searchv2/rules.py
```diff
@@ -132,6 +132,28 @@ def check(self, package: Package) -> CheckResult:
...
+    documentation_url: str = f"{DOCS_URL}/rules/favorite-package"
+
+    def check(self, package: Package) -> CheckResult:
+        """
+        Check if the package is favorite.
+        Returns a full score and a success message if the package is not favorite,
+        or a zero score and an error message otherwise.
+        """
+        if package.is_favorite:
+            return CheckResult(score=0, message="Package is favorite.")
```
saadmk11, line 152:
```suggestion
            return CheckResult(score=self.max_score, message="Package is favorite.")
```

searchv2/rules.py
```diff
@@ -132,6 +132,28 @@ def check(self, package: Package) -> CheckResult:
...
+        else:
+            return CheckResult(score=self.max_score, message="Package is not favorite.")
```
saadmk11, line 154:
```suggestion
            return CheckResult(score=0, message="Package is not favorite.")
```

package/models.py
```diff
@@ -158,6 +158,12 @@ def get_opengraph_image_url(self):
     def is_deprecated(self):
         return self.date_deprecated is not None
 
+    @property
+    def is_favorite(self):
```
saadmk11, line 162: What do you think about calling this `has_favorites`?

saadmk11 (changes requested):
Nice job 👍

package/models.py
lipemorais: Done  in d505400e3aadf30fbb6537183f5ac2be0398ed1a @saadmk11

searchv2/rules.py
lipemorais: Done f6421a52ec9320f4843e70c3891389c7ed3eb012

searchv2/rules.py
lipemorais: And done f6421a52ec9320f4843e70c3891389c7ed3eb012

jefftriplett (approved):
@lipemorais great work! We can work with this, and once we have more total favorites, we can update the check to tier the results to weight the point totals. Exciting!

PR #1247, opened 2024-08-07
:arrow_up: Upgrade Django to 5.1

### LINKED ISSUE #1246: :arrow_up: Upgrade to Django 5.1
[Django 5.1](https://www.djangoproject.com/weblog/2024/aug/07/django-51-released/) is out. 

And while we are in no rush to upgrade, we should see if anything breaks. In theory, we could target 5.1.1 or 5.1.2 unless everything just works. 

See also my guide on when I feel it's safe to upgrade: https://micro.webology.dev/2024/02/01/choosing-the-right.html 

Blocked on https://github.com/etianen/django-reversion/issues/974

jefftriplett:
Fixes #1246. This is mostly for testing. 

I think we could wait until Django 5.1.1 is out.


requirements.txt
```diff
@@ -347,9 +377,9 @@ distro==1.9.0 \
     # via
     #   cloudflare
     #   openai
-django==5.0.8 \
-    --hash=sha256:333a7988f7ca4bc14d360d3d8f6b793704517761ae3813b95432043daec22a45 \
-    --hash=sha256:ebe859c9da6fead9c9ee6dbfa4943b04f41342f4cea2c4d8c978ef0d10694f2b
+django==5.1 \
```
jefftriplett, line 380: Proof it worked.

jefftriplett:
We will need an upstream `django-reversion` release before this is ready.

PR #1255, opened 2024-08-21
Fix Large Grid Display Issue

saadmk11:
This PR adds overflow scrollbar for large grids, which used to break the UI.

**Example:** https://djangopackages.org/grids/g/admin-interface/

**Before:**
Screenshot from 2024-08-21 17-37-59

**After:**
Screenshot from 2024-08-21 17-37-33


jefftriplett (approved):
🎉  Wonderful update!

PR #1261, opened 2024-09-01
Migrate to `psycopg3` and add connection pooling

saadmk11:
Ref: https://docs.djangoproject.com/en/5.1/ref/databases/#connection-pool


jefftriplett:
This looks great. 

We are also using connection pooling on our DO-managed database. Have you seen any tips or details on using these both together? I'm also curious about the django-q2 workers and if we should bump our initial connection options.

saadmk11:
> We are also using connection pooling on our DO-managed database. Have you seen any tips or details on using these both together?

Oh! I was not aware of it. I think we can use both together but it reduces the necessity of application level connection pooling. I have seen people use both application level connection pooling and stand alone connection pooler like PgBouncer together for large applications. But this might be overkill for our use case.

> I'm also curious about the django-q2 workers and if we should bump our initial connection options.

We are using 4 workers for django-q2, and I am not sure if it is under heavy load constantly, if not max 4 connections should work okay.

jefftriplett:
@saadmk11 this is live now. 🚀

saadmk11:
@jefftriplett Awesome 🍻

I am curious about how this change will affect our production environment. Let me know if this breaks anything or you find something interesting in the logs/metrics. 🙂

dockerfiles/django/entrypoint.sh
```diff
@@ -6,10 +6,10 @@ cmd="$@"
 function postgres_ready(){
 python << END
 import sys
-import psycopg2
+import psycopg
 try:
-    conn = psycopg2.connect(dbname="$POSTGRES_DB", user="$POSTGRES_USER", password="$POSTGRES_PASSWORD", host="$POSTGRES_HOST", port="$POSTGRES_PORT")
-except psycopg2.OperationalError:
+    conn = psycopg.connect(dbname="$POSTGRES_DB", user="$POSTGRES_USER", password="$POSTGRES_PASSWORD", host="$POSTGRES_HOST", port="$POSTGRES_PORT")
```
pauloxnet, line 11: Here you could have used the `DATABASE_URL` variable.
https://www.psycopg.org/psycopg3/docs/api/connections.html#psycopg.Connection.connect

jefftriplett: We should remove this altogether. I just haven't gotten around to it yet. Thank you for bringing it up.

jefftriplett:
@saadmk11 I didn't catch this until Florian's mastodon post tipped me off. We reference the settings/docker.py in production so we were overriding connection pooling it looks like. 

https://mastodon.social/@apollo13@chaos.social/113160586859399474

jefftriplett:
> Nice Catch. 💯

Thank you! We should really merge everything into one settings file. I have meant to do that for a while to avoid this not being more obvious.

ryancheley:
> Thank you! We should really merge everything into one settings file. I have meant to do that for a while to avoid this not being more obvious.

@jefftriplett  Sounds like something we can sprint on next week 😁

jefftriplett:
> > Thank you! We should really merge everything into one settings file. I have meant to do that for a while to avoid this not being more obvious.
> 
> @jefftriplett Sounds like something we can sprint on next week 😁

Yes, plz!

Issue #915, opened 2023-02-11
⚙️ Consolidate settings file


jefftriplett:
Consider consolidating settings files to avoid needing one for docker and everything else.

jefftriplett:
@ryancheley this is live in production. Good work!

Issue #1276, opened 2024-09-26
Safari login broken in dev environment


giaume:
**Describe the bug**

Login is broken with Safari, works with Chrome and Firefox.

**To Reproduce**

- Install a fresh dev environment following README.md instructions
- Create a superuser
- Try to login from http://localhost:8000/admin/ using Safari
- It goes back to login/password without any error.

**Expected behavior**

Get logged in.

**Additional context**

Sniffing network shows the server response to POST has:

```
Set-Cookie:  sessionid=XXX; expires=Thu, 10 Oct 2024 18:59:02 GMT; HttpOnly; Max-Age=1209600; Path=/; SameSite=Lax; Secure
```

It looks like Safari is then not passing this cookie because it's http (not https) and cookie is marked as `Secure`,
therefore we get the login page again and no error message at all.

**Solution**

Change in `settings/docker.py`:

```
SESSION_COOKIE_SECURE = False
```

giaume:
Same behavior (can't log in) using Firefox or Chrome with http://0.0.0.0:8000/admin/

But for both Firefox and Chrome, it works with http://localhost/admin/ or http://127.0.0.1/admin/, they must have some hard-coded exceptions to ignore the `Secure` cookie attribute for `localhost` and for `127.0.0.1`, which Safari does not have.

jefftriplett:
This should now be taken care of in #1280. 

Thank you for reporting it and doing all of the work to track down a fix.

PR #1280, opened 2024-09-27
Ugly hack to start settings files consolidation

ryancheley:
References #915


ryancheley:
The tests are failing in CI, but the pass when run locally. I'm not sure why that would be. @jefftriplett @saadmk11 any ideas?

saadmk11:
> The tests are failing in CI, but the pass when run locally. I'm not sure why that would be. @jefftriplett @saadmk11 any ideas?

Probably because of this:

https://github.com/djangopackages/djangopackages/blob/7b3d078ec7722ff0627d1ceb0bdb0e7319f91831/pyproject.toml#L16

package/views.py
```diff
@@ -17,7 +17,7 @@
 from django.views.generic.base import TemplateView
 from django_q.tasks import async_task
 from django_tables2 import SingleTableView
-from settings.base import DOCS_URL
+from core.settings import DOCS_URL
```
jefftriplett, line 20: To avoid having to change this in the future, this should really be: 

```suggestion
from django.conf import DOCS_URL
```

searchv2/rules.py
```diff
@@ -3,7 +3,7 @@
 from pydantic import BaseModel
 
 from package.models import Package
-from settings.base import DOCS_URL
+from core.settings import DOCS_URL
```
jefftriplett, line 6:
```suggestion
from django.conf import DOCS_URL
```

jefftriplett (changes requested):
This isn't one I can deploy during nap time on a Sunday, but I think it's 98% there.

jefftriplett:
While we are here, let's fix `SESSION_COOKIE_SECURE` that's mentioned in #1276 so that local development works too.

core/settings.py
```diff
@@ -172,7 +211,19 @@
...
+    PREREQ_APPS
+    + PROJECT_APPS
+    + [
+        "anymail",
+    ]
+)
+
+ANYMAIL = {
+    "MAILGUN_API_KEY": env("MAILGUN_API_KEY"),
+    "MAILGUN_SENDER_DOMAIN": env("MAILGUN_SENDER_DOMAIN"),
```
saadmk11, line 224: I think we can add dummy defaults here

core/settings.py
```diff
@@ -1,9 +1,16 @@
 # Django settings
```
saadmk11, line 1: Could we put the settings file somewhere else instead of putting it inside a **Django App**?

saadmk11 (commented):
Awesome work here 💯 , changes look great. Added some small nits.

core/settings.py
```diff
@@ -172,7 +211,19 @@
     "template_partials",
 ]
 
-INSTALLED_APPS = PREREQ_APPS + PROJECT_APPS
+INSTALLED_APPS = (
+    PREREQ_APPS
+    + PROJECT_APPS
+    + [
+        "anymail",
```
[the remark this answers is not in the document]
ryancheley: @jefftriplett do you know why the `anymail` was split into it's own list? I'm good with moving it, just not sure why it would have been explicitly separated and want to make sure we don't make a change before knowing why it is the way it is 😄

core/settings.py
ryancheley: dummy values added

core/settings.py
ryancheley: @saadmk11 are you thinking at the top level with `manage.py` or something else?

jefftriplett: > @saadmk11 are you thinking at the top level with `manage.py` or something else?

I think at the top level is fine. I tend to put these in their own app (I prefer `config` over `djangopackage` but in this case the `settings/__init__.py` is roughly the same as `settings.py` from a namespace perspective. 

What I care the most about is that we are consistent and update everywhere this needs to be reflected in.

core/settings.py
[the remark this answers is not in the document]
jefftriplett: @ryancheley Nope, probably an oversight or it was shifted during a merge conflict/review.

ryancheley:
> While we are here, let's fix `SESSION_COOKIE_SECURE` that's mentioned in #1276 so that local development works too.

With the change I'm not able to replicate the issue 🤷

ryancheley:
@jefftriplett I think this is ready for another pass through. I think I've made all of the requested changes 😄

jefftriplett (approved):
Good work!

jefftriplett:
@ryancheley thank you for pushing this over the finish line. I will do a quick pr if something shakes out after merging. I see maybe an ENV var or two that we might not have to force to be required but nothing that should hold anything up.

Issue #1282, opened 2024-10-05
💅  Add djade


jefftriplett:
Since @ryancheley brought it up, [djade](https://github.com/adamchainz/djade) feels worth exploring. 

https://mastodon.social/@ryancheley/113256579926595513

saadmk11:
Interesting, I haven't heard of this project before. I added a PR (https://github.com/djangopackages/djangopackages/pull/1283) to try this out. :)

jefftriplett:
:rocket: This is deployed now.

PR #1283, opened 2024-10-06
Try Djade (Django template formatter)

saadmk11:
closes https://github.com/djangopackages/djangopackages/issues/1282

### DIFF
--- a/templates/_homepage.html
+++ b/templates/_homepage.html
@@ -1,9 +1,4 @@
-{% load emojificate %}
-{% load humanize %}
-{% load i18n %}
-{% load package_tags %}
-{% load typogrify_tags %}
-{% load waffle_tags %}
+{% load emojificate humanize i18n package_tags typogrify_tags waffle_tags %}
 
 
 {% switch "switch_use_grid_of_the_week" %}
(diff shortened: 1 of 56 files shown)


saadmk11:
Tests are passing :)

templates/_homepage.html
```diff
@@ -1,9 +1,4 @@
-{% load emojificate %}
-{% load humanize %}
-{% load i18n %}
-{% load package_tags %}
-{% load typogrify_tags %}
-{% load waffle_tags %}
+{% load emojificate humanize i18n package_tags typogrify_tags waffle_tags %}
```
saadmk11, line 1: I personally like each load tag on a separate line.

ryancheley: @saadmk11 I agree. There is a conversation in the [forum](https://forum.djangoproject.com/t/django-template-style-guide-extensions/35351) about this with some people wanting them on a single line alphabetized, with others advocating for a single line per library. 

There's also an [issue in djade](https://github.com/adamchainz/djade/issues/33) with a discussion about this very thing and @jefftriplett [indicating a preference for one load-statement per line](https://github.com/adamchainz/djade/issues/33#issuecomment-2386226216) because

> it streamlines code reviews and diffs.

It could be that options are added to `djade` to allow for this preference but that's still unclear

jefftriplett: Same, I think leaving a message on the post @ryancheley highlighted is a good place to mention that too. 

While I like that Black has only one configuration option, this is the one I hope `djade` adopts so that people have a choice.

jefftriplett (approved):
🎉  Looks good to me. I'm all for any linter that also auto-formats. We'll give it a try and see if anything gets weird.

saadmk11:
> 🎉  Looks good to me. I'm all for any linter that also auto-formats. We'll give it a try and see if anything gets weird. 

Awesome! I tested this locally but we should keep an eye on sentry just to be sure.

jefftriplett:
Clean for the last ~2 hours, but I will have it up for the night :)

Issue #1285, opened 2024-10-06
:tractor: Migrate to Python 3.12


jefftriplett:
Python 3.12 has been out long enough that it's safe to upgrade from 3.11 to 3.12. 

```shell
➜ git grep python3.1
.pre-commit-config.yaml:  python: python3.11

➜ git grep python:3
dockerfiles/django/Dockerfile:FROM python:3.11-slim-bookworm
dockerfiles/django/Dockerfile-dev:FROM python:3.11-slim-bookworm

➜ git grep py311
.pre-commit-config.yaml:        args: [--py311-plus]
pyproject.toml:target-version = "py311"

➜ git grep "3.11"
.devcontainer/Dockerfile:ARG VARIANT="3.11-bookworm"
.github/workflows/actions.yml:      - name: Set up Python 3.11
.github/workflows/actions.yml:          python-version: 3.11
.github/workflows/deploy-main.yml:      - name: Set up Python 3.11
.github/workflows/deploy-main.yml:          python-version: 3.11
.github/workflows/update-changelog.yaml:        python-version: 3.11
.github/workflows/update-contributors.yml:        python-version: 3.11
.pre-commit-config.yaml:  python: python3.11
.readthedocs.yml:    python: "3.11"
```

hriprsd:
I have modified to use python 3.12 and also did a blanket search across the repo to make sure its properly updated everywhere. 
Please lemme know if there's any more changes needed. Thanks for the opportunity! 😃

PR #1291, opened 2024-10-15
upgrading to python 3.12

hriprsd:
- upgrading to python 3.12
- solves #1285 
- python 3.12 used across project in the notified areas as in the issue. 
- updated changelog as well.


hriprsd:
Also, could you please add `hacktoberfest-accepted` label to my PR please? would be a huge help! Thanks in advance!

hriprsd:
Please do let me know if I have missed something/ any more changes are needed. Thanks in advance! 😄

hriprsd:
Hi. Hoping this will be counted towards my hacktober contribution. P.s would need the label `hacktoberfest-accepted` added to my PR or `hacktober` topic to the repo please. 

Aplogies for the follow ups, but within October would be great
 🙂 

Thanks in advance!

jefftriplett:
Thank you for your contribution. We'll review this week and add the tag after we accept it. Juggling day jobs but we will not forget to add it :)

pyproject.toml
```diff
@@ -63,7 +63,7 @@ line-length = 88
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 
 # Assume Python 3.11
-target-version = "py311"
```
saadmk11, line 66: please update the comment above this line (`# Assume Python 3.11`)

saadmk11 (changes requested):
Awesome work here 💯  This is looking great. 👍 

Seems like some tests are failing because python 3.12 removed `distutils` (`ModuleNotFoundError: No module named 'distutils'`), I think we can fix this by replacing `distutils` with the `packaging` package which we already install.

We may also want to re-compile the `requirements.txt` file with the new python version.
https://github.com/djangopackages/djangopackages/blob/c5fa68de85aa0c7e055ac4eac6090cf5f37d36ad/requirements.txt#L2

hriprsd:
> Awesome work here 💯 This is looking great. 👍
> 
> Seems like some tests are failing because python 3.12 removed `distutils` (`ModuleNotFoundError: No module named 'distutils'`), I think we can fix this by replacing `distutils` with the `packaging` package which we already install.
> 
> We may also want to re-compile the `requirements.txt` file with the new python version.
> 
> https://github.com/djangopackages/djangopackages/blob/c5fa68de85aa0c7e055ac4eac6090cf5f37d36ad/requirements.txt#L2

Fixed. 😄

jefftriplett:
TIL this, merged the PR for who only knows what reasons... 

```
$ git push origin hriprsd/main:main
```

jefftriplett:
@hriprsd thank you for the contribution

@saadmk11 @ryancheley This is fixed in the main, and I found a package that replaces our usage of LooseVersion.

Issue #1298, opened 2024-10-30
`install` link is broken


FarhanAliRaza:
**Describe the bug**
Install link is broken in github readme and docs `https://docs.djangopackages.org/en/latest/install.md` 

**To Reproduce**
1. Go to 'README.md'
2. Click on `docs` in the Quickstart section
3. See error

**Expected behavior**
Link broken

**Additional context**

PR #1299, opened 2024-10-30
broken link fix

FarhanAliRaza:
The `Installation` link was broken i updated it `contributing.md#contributing.md#install-django-packages-locally`
fixed https://github.com/djangopackages/djangopackages/issues/1298

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Django Packages helps you easily identify and compare good apps, frameworks, plu
 
 ## Quickstart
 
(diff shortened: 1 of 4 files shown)


saadmk11 (approved):
Nice work 🚀

FarhanAliRaza:
Now Opinionated Install and normal install is also merged into one so the both should like to the same url.

jefftriplett (approved):

jefftriplett:
@FarhanAliRaza nice catch!

PR #1302, opened 2024-11-02
:gear: :whale: Adds missing `--profile` option

jefftriplett:
This was tripping up a production cron issue that wasn't obvious. The `django-a` and `django-b` services were rebuilding, but the `utility` service wasn't getting a newer profile to rebuild.


jefftriplett:
Testing again.

saadmk11:
Nice catch

Issue #1266, opened 2024-09-10
:handshake: Support Codeberg


jefftriplett:
We are seeing some major projects start to move to Codeberg off of GitHub. I'm not sure what integration looks like but it's worth exploring.

One example: https://djangopackages.org/packages/p/django-allauth/

fallenhitokiri:
Codeberg seems to be running (a [slightly modified](https://codeberg.org/Codeberg-Infrastructure/forgejo) version of) [Forgejo](https://forgejo.org) for repository hosting.

The only difference to support self hosted Forgejo instances as well as Codeberg seems to be a little bit less strict regex for validating repository URLs.

If this turns out to be the case should a Codeberg repository handler inherit a Forgejo handler and add a more strict regex ensuring `codeberg.org` to be part of the URL or would a Forgejo repo handler be sufficient?

jefftriplett:
This worked locally for me with `django-allauth`, but it doesn't seem to pick up the stars and forks in production. I have looked a little, but I haven't spent a ton of time on it yet. 

https://djangopackages.org/packages/p/django-allauth/

jefftriplett:
Noted in https://github.com/djangopackages/djangopackages/issues/1324

fallenhitokiri:
For Forgejo support I think a few changes would be needed. From what I can tell

- `get_repo_for_repo_url` will not work as Forgejo instances can be self hosted and therefore no regex would match
- during package creation when the hosting provider cannot be determined there would need to be a way to indicate that the package is hosted on Forgejo
- the hosting option would be saved on the `Package`
- `get_repo_for_repo_url` would need to accept a package and return a handler based on the saved value

aayushgauba:
@joshuadavidthomas This seems interesting. I will be willing to take this on is if this still needs work. Thanks!

aayushgauba:
@joshuadavidthomas I got a initial version of things here #1435

PR #1321, opened 2024-12-26
codeberg support #1266

fallenhitokiri:
Adds Codeberg repo handler to fetch meta data and commits.

Issues:
- collaborators: API does not return any data (code and Swagger)

Feature: https://github.com/djangopackages/djangopackages/issues/1266


package/repos/codeberg.py
```diff
@@ -0,0 +1,190 @@
...
+        self.api_token = settings.CODEBERG_TOKEN
+
+    def fetch_commits(self, package: "package.models.Package"):
+        from package.models import Commit  # Added here to avoid circular imports
+
+        for commit in _fetch_commits(package.repo_name()):
+            try:
+                _, created = Commit.objects.get_or_create(
+                    package=package,
+                    commit_hash=commit.sha,
```
fallenhitokiri, line 51: I am using https://codeberg.org/allauth/django-allauth as test repository.

`GitHubHandler` only uses the package and the commit date to create commits. allauth has several commits at the same time with means the stop condition below triggers after 27 commits instead of doing an initial sync of all 3574 commits. Passing the commit hash fixes this problem.

Is there a reason this isn't needed for GitHub? (mili / nano second precession instead of seconds?)

package/repos/codeberg.py
```diff
@@ -0,0 +1,190 @@
...
+
+    data = response.json()
+
+    try:
+        meta = ForgejoMetadata(
+            archived=data["archived"],
+            archived_at=data["archived_at"],
+            description=data["description"],
+            forks_count=data["forks_count"],
+            watchers_count=data["watchers_count"],  # TODO: watchers or starts?
```
fallenhitokiri, line 106: This field is re-written to `Stars` in Django admin. Should this data be stars, watchers or stars + watchers?

package/repos/codeberg.py
```diff
@@ -0,0 +1,190 @@
...
+                break
+
+        if "x-hasmore" in response.headers and response.headers["x-hasmore"] == "true":
+            current_page = current_page + 1
+        else:
+            break
+
+
+def _fetch_collaborators(repository: str, page_size=50, token=""):
+    """TODO: right now this endpoint does not return any data."""
```
fallenhitokiri, line 152: I am looking into this, but even Swagger on Codeberg itself does not return any data for this endpoint.

package/repos/codeberg.py
jefftriplett: Great find. This code (outside of my lint update) is 9 to 11 years old. I suspect it was an oversight. 

I'll have to do some testing, but I think we could switch to `commit_hash` if you want to switch this new repo handler to it over `commit_date`.

package/repos/codeberg.py
jefftriplett: The original version was written with GitHub API's terminology, so let's keep the watchers (publicly called stars). (we might refactor this later)

package/repos/codeberg.py
jefftriplett: That's fine. I think we can add this later if Forgejo adds it. If we get enough and feel motivated, we can probably keep a checkout and scan them ourselves. (Let's punt for now.)

jefftriplett (changes requested):
This looks amazing! I left a follow-up comment and switching to the `commit_hash` which I am surprised I hadn't noticed that before. So fresh eyes are really appreciated.

package/repos/codeberg.py
saadmk11: Nice catch! 💯

package/repos/codeberg.py
fallenhitokiri: I could work around the endpoint by using the commit data to extract participant information.

It mixes repository meta data updates with commit updates. If we strictly want to separate these two it is likely not a good idea and I will revert the code.

ryancheley (approved):
This looks great. I tried it out locally and everything seems to work as expected. Great job on this @fallenhitokiri 💯 !

jefftriplett (approved):
Woot! 🎉

jefftriplett:
@fallenhitokiri great work! This was a huge help. Thank you!

Issue #1063, opened 2023-11-18
Integrate score value into Package/Grid results


ryancheley:
**Is your feature request related to a problem? Please describe.**
The current search results cant be swamped by historic stars on repositories making it difficult to find current packages with confidence

**Describe the solution you'd like**
Integrate score value into Package/Grid results

jefftriplett:
The scoring logic overlaps with #1062 or vice versa, depending on the order in which we pick them up.

aayushgauba:
@joshuadavidthomas @jefftriplett This seems interesting to me and I would like to work on this if this is still open

jefftriplett:
Closing - score values are integrated into package/grid results.

PR #1322, opened 2024-12-28
Integrate score value into Package/Grid results #1063

luanfonceca:
**Description:**
Adding the "Score" to both the Grid and Package detail pages.

**Issue:**
https://github.com/djangopackages/djangopackages/issues/1063

**Screenshots:**
Grid Detail page:

https://github.com/user-attachments/assets/f7d6c037-c71d-4b67-b3ef-7112f88e5dc9

Package Detail page:

https://github.com/user-attachments/assets/ecafd97d-0268-4bb4-80c0-31a80250dfc1

**Open questions:**
1. Should I use the `{% static <path-to-asset>%}` or the {{ STATIC_URL }}? I saw places where one or the other is being used.
2. I'm not the biggest fan of testing HTML content without test markup, such as a `data-test-id` attribute in the tag, to avoid false positives and flakiness when asserting that the content I expected is in the correct place. Do you have any objections to adding it to this PR?
3. Can we move the Bootstrap.js import to our `base.html` template? By doing that, we use the Tooltip markup whenever we want without needing to import it every time.

### DIFF
--- a/core/context_processors.py
+++ b/core/context_processors.py
@@ -38,3 +38,9 @@ def current_path(request):
     if request.path.strip() != reverse("logout"):
         context["current_path"] = request.path
     return context
+
+
+def settings_context(request):
+    return {
+        "TEST_MODE": settings.TEST_MODE,
+    }
--- a/grid/tests/test_views.py
+++ b/grid/tests/test_views.py
@@ -108,7 +108,7 @@ class FunctionalGridTest(TestCase):
             response,
             dedent(
                 """
-                    <td>
+                    <td data-testid="grid-detail-score-header">
                         Score
                         <span
                             class="glyphicon glyphicon-info-sign"
@@ -127,7 +127,7 @@ class FunctionalGridTest(TestCase):
                 response,
                 dedent(
                     f"""
-                        <td>
+                        <td data-testid="grid-{package.slug}-detail-score-cell">
                             {intcomma(package.score)}
                         </td>
                     """
--- a/grid/tests/test_views.py
+++ b/grid/tests/test_views.py
@@ -114,7 +114,7 @@ class FunctionalGridTest(TestCase):
                             class="glyphicon glyphicon-info-sign"
                             data-toggle="tooltip"
                             data-placement="top"
-                            title="Scores (0-100) are based on GitHub stars, with deductions for inactivity (-10% every 3 months) and lack of Python 3 support (-30%)."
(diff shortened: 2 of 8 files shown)


jefftriplett:
This is wonderful! 

> Should I use the {% static <path-to-asset>%} or the {{ STATIC_URL }}? I saw places where one or the other is being used.

Let's use `static`.  

> I'm not the biggest fan of testing HTML content without test markup, such as a data-test-id attribute in the tag, to avoid false positives and flakiness when asserting that the content I expected is in the correct place. Do you have any objections to adding it to this PR?

Nope, this makes sense to me. 

> Can we move the Bootstrap.js import to our base.html template? By doing that, we use the Tooltip markup whenever we want without needing to import it every time.

Sure thing. Adding Bootstrap.js is fine by me. (Side note: We are probably moving off of Bootstrap in the medium term, but we aren't there yet and will still have tooltips.) 

---- 

The one issue with exposing the score is that they might look wonky. See this screenshot: 


This might be good motivation for us to finish/clean-up/refine our new scoring system though.

luanfonceca:
> Sure thing. Adding Bootstrap.js is fine by me. (Side note: We are probably moving off of Bootstrap in the medium term, but we aren't there yet and will still have tooltips.)

Nice! I saw that the migration was started, but Tailwind itself doesn't have a built-in tooltip.
I have done my fair share of Bootstrap to Tailwind migrations in the past, so let me know if I can help with anything related to that area.

luanfonceca:
> This might be good motivation for us to finish/clean-up/refine our new scoring system, though.

Ohh, I should have looked at that `Package.calculate_score` logic further haha

IMO open-ended numbers don't make sense for a score because we don't have a ceiling value to base ourselves on and understand if that value I'm seeing is a good or a bad score.

I saw that we already have `django-waffle` installed, could I use a flag to hide this from the UI until we decide what to do with the scores?

jefftriplett:
> IMO open-ended numbers don't make sense for a score because we don't have a ceiling value to base ourselves on and understand if that value I'm seeing is a good or a bad score.

While trying to clean it up, I was running into data rabbit holes and a bunch of side requests like getting Pypi download stats, trying to find outliner packages that aren't really Django Packages (think popular Python packages like Black, Coverage, ...) which skew our numbers. I made some progress but never quite nailed it. 

> I saw that we already have `django-waffle` installed, could I use a flag to hide this from the UI until we decide what to do with the scores?

I think we can ship it without a flag. It'll be motivation to fix it 😬

jefftriplett (approved):
This looks great to me. Great work and we really appreciate your contribution!

Issue #1328, opened 2025-01-07
🐛 Codeberg participants can contain None failing the fetch_commits


fallenhitokiri:
**Describe the bug**
When synchronising a Codeberg repository collaborators can be `None` which causes the `join` to fail.

```python
>>> h.fetch_commits(p)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/code/package/repos/codeberg.py", line 147, in fetch_commits
    package.participants = ",".join(set(collaborators))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: sequence item 10: expected str instance, NoneType found
```

**To Reproduce**
1. Start the stack via docker compose
2. Get a shell
3. Synchronise a repository like `django-allauth`

**Expected behavior**
Have `Package.participans` populated after synchronising commits.

**Additional context**
Check for `None` before adding a name to the list https://github.com/djangopackages/djangopackages/blob/main/package/repos/codeberg.py#L143

hvdklauw:
Will this fix #1324?

jefftriplett:
I'm looking this afternoon. The feature works locally but not on our server which makes me wonder if we have something weird data wise.

jefftriplett:
I think @aayushgauba helped fix this. Status are flowing again https://djangopackages.org/packages/p/django-allauth/

jefftriplett:
@aayushgauba great work. I added https://djangopackages.org/packages/p/django-admin-deux/ tonight and it all just worked!

PR #1329, opened 2025-01-08
fix participant extraction in CodebergHandler

fallenhitokiri:
Fixes issue #1328

### DIFF
--- a/package/repos/codeberg.py
+++ b/package/repos/codeberg.py
@@ -140,11 +140,14 @@ class CodebergHandler(BaseHandler):
                 if not created:
                     break
 
-                collaborators.append(commit.user)
+                if commit.user and commit.user != "":
+                    collaborators.append(commit.user)
             except Commit.MultipleObjectsReturned:
                 continue
 
-        package.participants = ",".join(set(collaborators))
+        if len(collaborators) > 0:
+            package.participants = ",".join(set(collaborators))
+
         package.save()
 
         return package
--- a/package/repos/codeberg.py
+++ b/package/repos/codeberg.py
@@ -140,7 +140,7 @@ class CodebergHandler(BaseHandler):
                 if not created:
                     break
 
-                if commit.user and commit.user != "":
+                if commit.user:
                     collaborators.append(commit.user)
             except Commit.MultipleObjectsReturned:
                 continue


package/repos/codeberg.py
```diff
@@ -140,11 +140,14 @@ def fetch_commits(self, package):
                 if not created:
                     break
 
-                collaborators.append(commit.user)
+                if commit.user and commit.user != "":
```
saadmk11, line 143: Only checking `commit.user` should be enough right?

```suggestion
                if commit.user:
```

package/repos/codeberg.py
```diff
@@ -140,11 +140,14 @@ def fetch_commits(self, package):
...
+                    collaborators.append(commit.user)
             except Commit.MultipleObjectsReturned:
                 continue
 
-        package.participants = ",".join(set(collaborators))
+        if len(collaborators) > 0:
+            package.participants = ",".join(set(collaborators))
```
saadmk11, line 149: A `package` is only updated when there are `collaborators` present, so, I think we should only call `save()` if that condition is `True`

```suggestion
        if collaborators:
            package.participants = ",".join(set(collaborators))
            package.save()
```

saadmk11 (commented):
Thanks for the PR 🚀 , changes look good, requested some small changes.

package/repos/codeberg.py
fallenhitokiri: I assumed syncing commits falls under the notion of a "repository fetch" and `Package.last_fetched` should be updated, even if participants don't change?

jefftriplett:
I'll run through this tomorrow (Friday afternoon, my timezone). Looks great so far.

package/repos/codeberg.py
```diff
@@ -140,11 +140,14 @@ def fetch_commits(self, package):
...
-                collaborators.append(commit.user)
+                if commit.user:
+                    collaborators.append(commit.user)
             except Commit.MultipleObjectsReturned:
                 continue
 
-        package.participants = ",".join(set(collaborators))
+        if len(collaborators) > 0:
+            package.participants = ",".join(set(collaborators))
+
```
pauloxnet, line 150: Maybe you can add `updated_fields` to speed up the save process.

pauloxnet (commented):
Maybe a bulk create or update will speed up the fetch process

package/repos/codeberg.py
jefftriplett: `last_fetched` should be updated per run. We use it to order the Packages when we are fetching for updates which solved an issue where some packages in the list where never getting updated because of another bug.

Issue #1062, opened 2023-11-18
Integrate score value into Search results


ryancheley:
**Is your feature request related to a problem? Please describe.**
The current search results cant be swamped by historic stars on repositories making it difficult to find current packages with confidence

**Describe the solution you'd like**
Integrate the updated scoring system into the Search results

- [ ] Ensure tests are created

ryancheley:
@jefftriplett found [this](https://jsr.io/@doctor/iterstar/score) and it looks like a potential idea for integration and display for the scores and the rules behind the scores.

ryancheley:
@luanfonceca that sounds like a great idea 💯. Please go ahead with that plan and thank you!

jefftriplett:
🚀  This is live. Great work @luanfonceca

luanfonceca:
> 🚀  This is live. Great work @luanfonceca 
> 
> 

Thanks, Jeff!

I was hoping to start defining what we want to change in the score calculation in the following weeks, but I'm afraid I won't have time to do that.

So, if you want to define how and what we should use to calculate the score, I believe I can eventually work on it.

PR #1333, opened 2025-01-14
#1062 Integrate score value into Search results

luanfonceca:
**Fixes:**
https://github.com/djangopackages/djangopackages/issues/1062

**Description:**
The Score is added to the Results page only for packages.

Changes:
1. the `LocMem` cache backend was added to avoid unnecessary database hits during tests. See commit 0ca463e0dafe64c1c891d66c6c72985adb9f2864.
2. Added a `score` field to the `SearchV2` model.
3. Applied the same visuals to the Search results found on the Package Detail page from PR #1322 

**Screenshots:**

DevTools localhost 8000 (8)
DevTools localhost 8000 (7)
DevTools localhost 8000 (6)
DevTools localhost 8000 (5)
DevTools localhost 8000 (4)

**How to Reproduce:**
**Scenario 1**:
With the `enabled_packages_score_values` flag **disabled**

1. Open the http://localhost:8000/search/?q=django with a Django search
4. Make sure that the Score is not visible in the table

**Scenario 2**:
With the `enabled_packages_score_values` flag **enabled**

1. Go to the Admin, flags page: http://localhost:8000/admin/waffle/flag/
2. Enable/Create the `enabled_packages_score_values` feature flag
5. Open the http://localhost:8000/search/?q=django with a Django search
6. Make sure that the Score is now visible in the table
7. Open the http://localhost:8000/search/?q=testing with the "testing" search
8. Make sure that the Score will be `N/A` for Grids, but still visible for Packages


ryancheley:
@luanfonceca this looks great! Thanks for the steps to reproduce. Very helpful. 

One thing I noticed is that in Chrome the bottom line doesn't extend to `Forks`. It does in Safari and Vivaldi though 🤔 

Screenshot 2025-01-18 at 6 55 23 AM

luanfonceca:
> @luanfonceca this looks great! Thanks for the steps to reproduce. Very helpful.
> 
> One thing I noticed is that in Chrome the bottom line doesn't extend to `Forks`. It does in Safari and Vivaldi though 🤔
> 
> Screenshot 2025-01-18 at 6 55 23 AM

Good catch! I'm using Arc and Safari so that I couldn't spot that one.

I found that I had the wrong variable in the templates used by Angular, and also, there's no `ng-attr` directive in Angular v1.1.5 - so I fixed that. I don't believe that would be the root cause here, but it's not good to have it there anyways.


Could you give it another try? I tried on a Chrome 132.0.6834.84 (arm64) within macOs but still can't reproduce.

I have a feeling that it could be related to things rendered by Django versus those rendered by Angular on the client side, so here's a diff patch you could apply that makes the Score table header to be also rendered by Angular:
```diff
diff --git a/templates/base.html b/templates/base.html
index 2eb3a9d8..91c2148b 100644
--- a/templates/base.html
+++ b/templates/base.html
@@ -120,20 +120,19 @@
                                 <th>{% translate "Package" %}</th>
                                 <th>{% translate "Description" %}</th>
                                 <th>{% translate "Last PyPI release" %}</th>
-                                {% flag "enabled_packages_score_values" %}
-                                    <th
-                                        {% if TEST_MODE %}data-testid="search-results-score-header"{% endif %}
-                                        scope="col"
-                                        data-toggle="tooltip"
-                                        data-placement="bottom"
-                                        container="body"
-                                        aria-label="{% translate 'Score' %}"
-                                        title="{% translate 'Scores (0-100) are based on Repository stars, with deductions for inactivity (-10% every 3 months) and lack of Python 3 support (-30%).' %}"
-                                    >
-                                        {% translate "Score" %}
-                                        <span class="glyphicon glyphicon-stats"></span>
-                                    </th>
-                                {% endflag %}
+                                <th
+                                    {% if TEST_MODE %}data-testId="search-results-score-header"{% endif %}
+                                    ng-if="waffle_flag_is_active('enabled_packages_score_values')"
+                                    scope="col"
+                                    data-toggle="tooltip"
+                                    data-placement="bottom"
+                                    container="body"
+                                    aria-label="{% translate 'Score' %}"
+                                    title="{% translate 'Scores (0-100) are based on Repository stars, with deductions for inactivity (-10% every 3 months) and lack of Python 3 support (-30%).' %}"
+                                >
+                                    {% translate "Score" %}
+                                    <span class="glyphicon glyphicon-stats"></span>
+                                </th>
                                 <th
                                     scope="col"
                                     data-toggle="tooltip"

```

You can apply it with `pbpaste | patch -p1 $(xargs)` on a macOS, and, I think, `xclip` on UNIX.

urls.py
```diff
@@ -11,19 +11,20 @@
...
+from homepage.views import (
+    OpenView,
+    ReadinessDetailView,
+    ReadinessView,
+    error_404_view,
+    error_500_view,
+    error_503_view,
+    health_check_view,
+    homepage,
+)
```
jefftriplett, line 23: Overall, everything looks great. This is my only nitpick.

conftest.py
```diff
@@ -17,13 +16,24 @@
     "CACHES": {
         "default": {
             "BACKEND": "django.core.cache.backends.dummy.DummyCache",
-            # "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
-        }
+        },
+        "waffle_cache_backend": {
+            "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+            "LOCATION": "unique-snowflake",
+        },
```
jefftriplett, line 23: Good catch! 👍

jefftriplett (approved):
Looks good! Thank you for the contribution.

Issue #1324, opened 2024-12-31
🐛 CodeBerg isn't refreshing stars and forks


jefftriplett:
Url: https://djangopackages.org/packages/p/django-allauth/

Hitting refresh isn't finding star and fork counts. We also aren't seeing commits from the graph but I think that was to be expected. 

<img width="846" alt="Image" src="https://github.com/user-attachments/assets/f1a75909-1ba0-48d4-9a5b-4fa5c4eb0e3e" />

jefftriplett:
This strangely might be some combo of GitHub and Codeberg data/streaming crossing.

fallenhitokiri:
I think I was able to reproduce this locally.

Starts and forks should show up when `Codeberg.fetch_metadata` is called.
The commit graph started to work as expected when I wiped all old commits before calling `Codeberg.fetch_commits`<img width="882" alt="Image" src="https://github.com/user-attachments/assets/0ccb4486-8204-44e7-aa63-4c41351a48d5" />

saadmk11:
I tested this locally, and it is working as expected. Here are the two variations I tried:

1. Added the package with **GitHub** URL, fetched the latest data, then updated the URL to **Codeberg** and fetched the latest data again.
2. Deleted the package via the Django admin, re-added it with **Codeberg** URL, and fetched the latest data.

From the package detail page I can see that `Package.last_fetched` is updated every time `Fetch latest data` button is clicked, which means `django-q` ran the task task successfully.

@jefftriplett did you see any Sentry issues that might be related to this bug?

jefftriplett:
> I tested this locally, and it is working as expected. Here are the two variations I tried:
> 
> 1. Added the package with **GitHub** URL, fetched the latest data, then updated the URL to **Codeberg** and fetched the latest data again.
> 2. Deleted the package via the Django admin, re-added it with **Codeberg** URL, and fetched the latest data.
> 
> From the package detail page I can see that `Package.last_fetched` is updated every time `Fetch latest data` button is clicked, which means `django-q` ran the task task successfully.
> 
> [@jefftriplett](https://github.com/jefftriplett) did you see any Sentry issues that might be related to this bug?

Sentry is clear outside of a random Redis connection timeout that happens once a week (seems normal) and an error every few days about an unrelated package whose score is zero. (I'll write that up)

jefftriplett:
We fixed this on Friday. Some of the container images were out of sync and was running outdated services for celery and utility services. This impacted search and some other processes, too.

PR #1340, opened 2025-01-28
Cleanup `SUPPORTED_REPO` Settings

saadmk11:


ryancheley:
@saadmk11 is this related to the bug in #1324

settings.py
```diff
@@ -313,9 +312,7 @@
 if LOCAL_INSTALLED_APPS:
     INSTALLED_APPS.extend(LOCAL_INSTALLED_APPS)
 
-# SUPPORTED_REPO.extend(["bitbucket", "github", "gitlab"])
-SUPPORTED_REPO.extend(["bitbucket", "github", "codeberg"])
-
+SUPPORTED_REPO = ["bitbucket", "github", "codeberg"]
```
jefftriplett, line 315: I believe we need GitLab too. 

```suggestion
SUPPORTED_REPO = ["bitbucket", "github", "gitlab", "codeberg"]
```

saadmk11: Interesting, why was gitlab not enabled before?

saadmk11:
> @saadmk11 is this related to the bug in #1324 

I don't think so.

settings.py
jefftriplett: Hmm, good point. I'll have to local test it.

jefftriplett (approved):
LGTM. thanks for catching it. It might be a bit before I can double check locally and then deploy it. (this afternoon?)

Issue #1386, opened 2025-05-03
:recycle: Clean Up justfile


ryancheley:
Based on the work in #1383 the justfile there are some commands which are helpful for the server admin and aren't usable by a non-server admin. I think these should be called out more specifically and added to a `[server-admin]` group. 

For example, the recipe `deploy` require the server root password and is in the `[deployment]` group. A second group could be added to this to make it clearer that only a server admin would be able to use it. 

There are also some commands which require a virtual environment, so adding them to a group called something like `[venv-required]` might be helpful ... or adding to the doc string that a virtual environment is required. 

We can also add the following command so that it will run on Windows. This does not impact the ability to run on macOS

```
set windows-shell := ["powershell.exe", "-NoLogo", "-Command"]
```

jefftriplett:
I can't speak for the Windows side (that can be in it's own ticket via #1383), but the `utils` tasks can probably all be refactored to use `uv tool run` and the need for a venv can go away. 

here are some quick thoughts: 

- [ ] `experimental` can go away
- [ ] We can lose all/most the `tailwind` commands once we have a tailwind watch command
- [ ] The `pre-commit*` commands can just be part of `lint` (and move it to `prek`) and then section can get absorbed into our defaults. 
- `deployment` can be renamed into `production`
- `server-admin` can be merged into `production`
- `testing` can lose pytest-* and `pytest-coverage` can be just `coverage`
- `docs` can be absorbed into a `compose` profile. 
- `management-command` can be renamed into `run`

jefftriplett:
Closing - justfile has been cleaned up and organized.

PR #1387, opened 2025-05-03
Clean up justfile

ryancheley:
Refs #1386


ryancheley:
@jefftriplett there were a few things that I found that I'm not sure how to resolve:

1. The tailwind commands all fail because of a missing file `./static/js/tailwind.config.js`
2. The `cron` recipe fails because there isn't a management command called `packages_download_stats`
3. The `test` recipe fails because

> The Django Debug Toolbar can't be used with tests
> 
> HINT: Django changes the DEBUG setting to False when running tests. By default the Django Debug Toolbar is installed because DEBUG is set to True. For most cases, you need to avoid installing the toolbar when running tests. If you feel this check is in error, you can set `DEBUG_TOOLBAR_CONFIG['IS_RUNNING_TESTS'] = False` to bypass this check.

I'm not sure how to resolve this. Perhaps we just remove the `test` recipe since we're using pytest

jefftriplett:
Sorry, this fell off my radar. I'll look when I get a sec.

justfile
```diff
@@ -39,7 +42,16 @@ bootstrap *ARGS:
     #!/usr/bin/env bash
     set -euo pipefail
 
-    python -m pip install --upgrade pip uv
+    # Check if virtual environment exists (looking for bin/activate)
+    if [ -d ".venv" ] && [ -f ".venv/bin/activate" ]; then
```
jefftriplett, line 46: Can we look for `VIRTUAL_ENV`  or some other variable that indicates that we are in an activated venv? 

I use virtualenv-wrapper and don't have an .venv folder in my projects.

jefftriplett (commented):
Looks good. There's one note about `VIRTUAL_ENV` which I think we just need to test it's behavior but it'll do what you want.

ryancheley:
@jefftriplett I think I've made an update that addresses the VIRTUAL_ENV issue. it does assume that uv will be used to create the virtual environment, but 🤷🏻‍♂️ ... kind of seems the way things are going

ryancheley:
Not sure why, but the tests pass locally and not in CI. The failing test is

`homepage/tests/test_views.py`

jefftriplett:
@ryancheley apparently this just needed to sit for ~2 months to air out. 

I'm going to merge it, but I'm tempted to adopt Hynek's Bringing Order to Convenience's convention for group names: https://www.youtube.com/watch?v=TiBIjouDGuI&t=1771s

ryancheley:
> ...  I'm tempted to adopt Hynek's Bringing Order to Convenience's convention for group names: [youtube.com/watch?v=TiBIjouDGuI&t=1771s](https://www.youtube.com/watch?v=TiBIjouDGuI&t=1771s)

I'm only about 10 minutes into the video and am really intrigued. His videos have just been so good ... really hoping I can wrap it up tonight

jefftriplett:
@ryancheley check out the group() labels here: https://github.com/hynek/hello-svc-part-2/blob/main/Justfile

PR #1392, opened 2025-05-27
Fix Homepage Test

saadmk11:
This PR https://github.com/djangopackages/djangopackages/pull/1330 introduced one new SQL query to the home page, previous to this the SQL query count was 11 and now it should be 12.

https://github.com/djangopackages/djangopackages/blob/31eafec164cba750bb8fb66fdd3c5b24450fd7ea/homepage/views.py#L273-L279


ryancheley (approved):
looks good to me

PR #1434, opened 2025-10-24
No need to say 'Python 3' on the homepage in 2025

hugovk:


jefftriplett (approved):
Looks good. thank you!

PR #1435, opened 2025-10-26
Forgejo changes

### LINKED ISSUE #1266: :handshake: Support Codeberg

aayushgauba:
Initial working version to Fix #1266 
@joshuadavidthomas Lemme know if this looks good or if there are things missing in here feature wise


core/utils.py
```diff
@@ -34,22 +35,24 @@ def get_pypi_url(title: str, timeout: float = 1.0):
...
+    PLANNING = 1, "Development Status :: 1 - Planning"
+    PRE_ALPHA = 2, "Development Status :: 2 - Pre-Alpha"
+    ALPHA = 3, "Development Status :: 3 - Alpha"
+    BETA = 4, "Development Status :: 4 - Beta"
+    STABLE = 5, "Development Status :: 5 - Production/Stable"
+    MATURE = 6, "Development Status :: 6 - Mature"
+    INACTIVE = 7, "Development Status :: 7 - Inactive"
+
+
+STATUS_CHOICES = PackageStatus.choices
```
joshuadavidthomas: Since this is only used in one spot, the following function, I would just remove this and use `PackageStatus.choices` directly.

package/tests/test_repos.py
```diff
@@ -59,6 +60,7 @@ def test_base_handler_get_repo_for_repo_url():
 http://code.google.com/p/django-courier
 http://django-cube.googlecode.com/hg
 http://launchpad.net/django-debian
+
```
joshuadavidthomas: Extra newline here, could be cleaned up:


```suggestion
```

package/tests/test_repos_codeberg.py
```diff
@@ -15,12 +11,18 @@ class MockClient(ForgejoClient):
     participants and archiving packages.
     """
 
+    def __init__(self):
+        # Avoid calling the parent constructor which requires a base_url argument
+        self.base_url = "https://codeberg.org"
+        self.api_base_url = f"{self.base_url}/api/v1"
```
joshuadavidthomas: No need to avoid calling the parent `__init__` method, you could just pass in the `base_url` to it and avoid the small duplication:

```suggestion
    def __init__(self):
        super().__init__(base_url="https://codeberg.org")
```

package/repos/forgejo.py
```diff
@@ -0,0 +1,209 @@
...
+            response = httpx.get(url)
+        except Exception as exp:
+            logger.error(exp)
+            return None
+
+        if response.status_code != httpx.codes.OK:
+            logger.error("Response code: %s for URL %s", response.status_code, url)
+            return None
+
+        data = response.json()
```
joshuadavidthomas: A more concise and idiomatic approach here would be to use `raise_for_status`:

```suggestion
        try:
            response = httpx.get(url)
            response.raise_for_status()
            data = response.json()

        except (httpx.HTTPError, ValueError) as exc:
            logger.error("Failed to fetch %s: %s", url, exc)
            return None
```

This should catch all the httpx transport and general request errors via `httpx.HTTPError` while allowing more general errors to bubble up by now swallowing via a generic `Exception` catch.

jefftriplett (changes requested):
I'm going to want to see only one migration file. I'd run something like: 

```shell
python manage.py migrate package 0021
```

Then delete your 0022 + 0023 migration files and run a clean make migrations.

aayushgauba:
@jefftriplett @joshuadavidthomas Fixed the issues you guys highlighted pointed out and let me know if I am missing something

jefftriplett:
I pushed a fix for your test (which I expected to fail) but then GH merged it from the cli.

Working on a fix.

jefftriplett:
@aayushgauba let's continue from #1441

jefftriplett:
@joshuadavidthomas @aayushgauba I had a series of `gh` failures, so this made it in before either of us intended. I think when I did a git push to the branch, it caught the `:main` part and merged it which is one hell of a bug. 

We should verify this and create an issue from anything else that comes from it.

PR #1442, opened 2025-11-06
Forgejo test Updates

aayushgauba:
Updated the Forgejo test to assert against the Forgejo API’s watchers_count


aayushgauba:
@jefftriplett This should fix the failing test. Updated the Forgejo test to assert against the Forgejo API’s watchers_count so it now matches the handler change that persists watchers instead of stars.

jefftriplett:
This took a while to track down, we really need to change:

```diff
diff --git a/package/repos/forgejo.py b/package/repos/forgejo.py
index f86b4657..ccf126f2 100644
--- a/package/repos/forgejo.py
+++ b/package/repos/forgejo.py
@@ -187,7 +187,8 @@ class ForgejoHandler(BaseHandler):
 
         package.repo_description = repo.description
         package.repo_forks = repo.forks_count
-        package.repo_watchers = repo.stars_count
+        package.repo_watchers = repo.watchers_count
+        # package.stargazers_count = repo.stars_count
         package.save()
 
         return package
```

tl;dr GitHub's API makes this even more confusing with `stargazers_count` which I believe are the number of watchers we have. We might want to add that too at some point and re-key everything. 

More context: https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository

aayushgauba:
@jefftriplett Should be good for now I think

jefftriplett:
@aayushgauba we were missing one more change that I picked up in #1444 

Good work on this. Let's double check the non-test code is doing the right things, but I think we are landing in the right palce.

aayushgauba:
> @aayushgauba we were missing one more change that I picked up in #1444
> 
> Good work on this. Let's double check the non-test code is doing the right things, but I think we are landing in the right palce.

Checked everything, and no other non-test references still use stars_count, and the rest of the handler logic (archived handling, forks, description) remains untouched. Theoretically, it should be good, but I guess we will find out

PR #1453, opened 2025-11-12
Mobile friendly changes

### LINKED ISSUE #1126: :children_crossing: Make main grid of categories on homepage more mobile friendly
**Is your feature request related to a problem? Please describe.**
I find the styling of the main grid of categories on the homepage a bit frustrating to use on mobile. The tap targets of the buttons on the side to scroll the list are so small and right up against the edge of the viewport. This wouldn't be so much of a problem if the list responded to swipe gestures, but it doesn't.

**Describe the solution you'd like**
1. Increase the padding of the buttons slightly to increase the tap target.
2. Make the grid swipeable. IMO, the simplest and modern solution would be to use CSS and scroll snap ([MDN Ref](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scroll_snap)), however that is complicated by the fact the current behavior is using jQuery to get the scroll behavior. I'm sure there's a way to get the current jQuery implementation to take swipe gestures, but it may be worth it to invest the time to rip the jQuery out and convert completely over to the CSS scroll snap approach. Or maybe a mix of the two? 🤷‍♂️

**Describe alternatives you've considered**
If the search was somehow visible on the mobile layout, I suppose that would alleviate the issue of trying to find a certain category on mobile, but that seems a bit orthogonal to the core of the issue.

**Additional context**
320063609-8ae2d655-6a82-4052-8784-918cb8116fa9 (Phone)

aayushgauba:
Mobile Friendly changes
Fixes #1126


joshuadavidthomas:
Yeah, this is a good start, but I think it's a *bit* too much of a radical change from the current styling.

image

That's on me for having such an open-ended issue. I think on desktop, the grid styling should largely be unchanged:

image

I think for mobile, you had the right idea, but I think it would be better instead of the next/prev buttons being on top and bottom, to just put them on the bottom:

image
image

And the grids should be fixed to only show on column per screen.

I will say, on mobile, the scroll snap on swipe is *chef's kiss* great. We just need to work on keeping the grid styling the same, IMO.

These are just my first thoughts, and the other maintainers may disagree. In my head, the styling should mostly be unchanged, except on mobile making slight tweaks to make it fit better in the smaller viewport. And the scroll snap behavior, which you've already nailed.

aayushgauba:
> Yeah, this is a good start, but I think it's a _bit_ too much of a radical change from the current styling.
> 
> image
> That's on me for having such an open-ended issue. I think on desktop, the grid styling should largely be unchanged:
> 
> image
> I think for mobile, you had the right idea, but I think it would be better instead of the next/prev buttons being on top and bottom, to just put them on the bottom:
> 
> image image
> And the grids should be fixed to only show on column per screen.
> 
> I will say, on mobile, the scroll snap on swipe is _chef's kiss_ great. We just need to work on keeping the grid styling the same, IMO.
> 
> These are just my first thoughts, and the other maintainers may disagree. In my head, the styling should mostly be unchanged, except on mobile making slight tweaks to make it fit better in the smaller viewport. And the scroll snap behavior, which you've already nailed.



> Yeah, this is a good start, but I think it's a _bit_ too much of a radical change from the current styling.
> 
> image
> That's on me for having such an open-ended issue. I think on desktop, the grid styling should largely be unchanged:
> 
> image
> I think for mobile, you had the right idea, but I think it would be better instead of the next/prev buttons being on top and bottom, to just put them on the bottom:
> 
> image image
> And the grids should be fixed to only show on column per screen.
> 
> I will say, on mobile, the scroll snap on swipe is _chef's kiss_ great. We just need to work on keeping the grid styling the same, IMO.
> 
> These are just my first thoughts, and the other maintainers may disagree. In my head, the styling should mostly be unchanged, except on mobile making slight tweaks to make it fit better in the smaller viewport. And the scroll snap behavior, which you've already nailed.

Hey @joshuadavidthomas, what do you think of this? I made a few changes that should take care of it.

PR #1465, opened 2025-12-05
Dark mode

tnware:
image

image

image

* First, it checks if you’ve clicked the theme toggle before.
  * If `localStorage` has `"dark"`, it forces Dark Mode.
  * If `localStorage` has `"light"`, it forces Light Mode.

* If there’s no saved preference in `localStorage`, it reads your OS/browser `prefers-color-scheme`.

* If there’s no saved preference and no system preference is available, it defaults to Light Mode.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -22,7 +22,7 @@
         {% block extra_head %}{% endblock %}
     </head>
     <body
-        class="flex flex-col min-h-screen dark:bg-gray-900 dark:text-gray-100"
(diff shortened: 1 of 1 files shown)


tnware:
this ended up being messy and relatively incorrect, closing this for now.
the tailwind + bootstrap situation isn't helped by adding this

jefftriplett:
@tnware thanks for attempting it. I'd like to dump bootstrap in favor of Tailwind and then it should be an easier lift.

Issue #1445, opened 2025-11-08
🚜  Migrating Server Public Issue for Feedback


jefftriplett:
We are moving our servers from Digital Ocean over to Hetzner. If you are seeing the banner, you are already on the new servers. 

We will use this issue to track updates. 

## DNS Changes 
- [x] https://djangopackages.org
- [x] https://www.djangopackages.org
- [x] https://node6.djangopackages.org
- [x] https://djangopackages.com
- [x] https://www.djangopackages.com
- [x] https://node6.djangopackages.com

## Related Issues 
- [x] Related to #1436 
- [ ] Related to #1433

jefftriplett:
Fixed a login redirect issue.

jefftriplett:
@ryancheley @saadmk11 @joshuadavidthomas DNS is updated and the website looks good so far. There are a few slow spots, but I'll keep an eye on it for a bit tonight. 

There's more on the backend to do for cron jobs, etc but this seems promising so far.

jefftriplett:
I think we can take the banner off. Things seemed baked and stable.

PR #1473, opened 2025-12-25
Remove maintenance alert

saadmk11:
I did not remove the code as we might need this in the future for other notices/alerts.

clases https://github.com/djangopackages/djangopackages/issues/1445

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -34,9 +34,9 @@
             </div>
         {% endfor %}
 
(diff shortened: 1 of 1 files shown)


jefftriplett:
I can update the message. I left it up because I haven't moved the database off of DO yet.

jefftriplett:
DB still needs removed but I'm in a cleanup the PR mood today. 

Good work!

PR #1483, opened 2026-01-10
Fix Search Suggestion Taps on Mobile Devices

saadmk11:
After https://dev.djangopackages.org/ was deployed I tried the new theme on a mobile device and found out that tapping on the search suggestion results (both for home page and navbar) on mobile was not doing anything (expected to go to the package/grid detail page), but this works on mobile view on the browser (which I used to test while developing). This PR fixes the Issue.

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -258,69 +258,93 @@ document.addEventListener("DOMContentLoaded", function () {
     });
 
     // ============================================
-    // Search Suggestions Keyboard Navigation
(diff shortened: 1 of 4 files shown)


ryancheley:
Not sure if we're able to (yet!), but this would be a really good use case for Preview deployments

jefftriplett:
No previews yet, but this will deploy to the dev website. We can turn on the preview deployments once we get the Team/Resources moved.

PR #1494, opened 2026-01-13
:hammer: Fix apiv3 n plus 1 queries

jefftriplett:


saadmk11 (approved):
Nice updates 💯

package/views.py
```diff
@@ -604,6 +612,13 @@ def get_base_queryset(self):
         return (
             Package.objects.active()
             .select_related("category")
+            .prefetch_related(
+                Prefetch(
+                    "version_set",
+                    queryset=Version.objects.only("package_id", "number"),
```
saadmk11: @jefftriplett  We should add .order_by("-upload_time"), see get_version() method.

PR #1495, opened 2026-01-13
:racehorse: Cache grid detail and invalidate on edits

jefftriplett:
This was very quickly generated with Codex before I called it a night.


saadmk11:
I also have a slightly different implementation I was working on yesterday

saadmk11:
This is my implementation: https://github.com/djangopackages/djangopackages/pull/1496

Issue #1500, opened 2026-01-13
Disable features for grids with more than 8 packages


jefftriplett:
## Summary

For performance reasons, we've disabled feature comparison on grids that have more than 8 active packages. This prevents expensive queries for features and elements on large grids.

## Changes

**Branch:** `disable-features-for-large-grids`

### GridDetailView (`grid/views.py`)
- Added `show_features` boolean that is `True` only when `total_package_count <= max_packages` (8)
- When `show_features` is `False`, the view skips querying features and elements entirely
- Passes `show_features` to templates via context

### Templates
- `templates/grid/grid_detail.html`: 
  - Features Being Compared card now checks `{% if show_features and features %}`
  - Timesheet button commented out (has no package limit)
- `templates/partials/grid_comparison_table.html`:
  - Desktop feature comparison rows check `{% if show_features and features %}`
  - Mobile feature comparison section checks `{% if show_features and features %}`

## Related

- #1498 - Commits sparkgraphs disabled for performance

PR #1501, opened 2026-01-13
Disable features for grids with more than 8 packages

jefftriplett:
## Summary

Disables feature comparison on grids with more than 8 active packages to improve performance. When disabled, the view skips querying features and elements entirely.

## Changes

- Add `show_features` boolean to GridDetailView (True when total_package_count <= 8)
- Skip features/elements queries when `show_features` is False
- Update templates to check `show_features and features`
- Disable Timesheet button (has no package limit)

## Test plan

- [ ] Verify grids with 8 or fewer packages still show features
- [ ] Verify grids with more than 8 packages do not show features
- [ ] Verify Timesheet button is hidden

See #1500

### DIFF
--- a/grid/views.py
+++ b/grid/views.py
@@ -298,7 +298,9 @@ class GridDetailView(DetailView):
 
         if show_features:
             features = self._get_features(grid)
-            elements = self._get_elements(features=features, grid_packages=grid_packages)
+            elements = self._get_elements(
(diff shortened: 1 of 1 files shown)


jefftriplett:
cc @saadmk11 we can turn these back on once we are stable.

PR #1507, opened 2026-01-14
[Part 1] Move Frequently Used Stats to Package Model and Refactor PyPI Fetch Functionality

saadmk11:
As we have disabled most of the related changes we can go through with this without breaking the site.

- [x] Management command to sync the new Package Fields
- [ ] Update the codebase to use the new package fields [Part 2 -- in a different PR]

Added new fields:
```python
    commits_over_52 = models.TextField(_("Commit List Over 52 Weeks"), blank=True)
    last_commit_date = models.DateTimeField(blank=True, null=True)
    latest_version = models.ForeignKey(
        "Version",
        blank=True,
        null=True,
        on_delete=models.SET_NULL,
        related_name="latest_version_package",
    )
```

These are used everywhere and currently we need to perform complex queries or use caching to get these values which slows down many queries.

@jefftriplett @ryancheley Any thoughts?

### DIFF
--- a/package/tests/test_sync_package_fields.py
+++ b/package/tests/test_sync_package_fields.py
@@ -40,7 +40,7 @@ def test_sync_package_fields_updates_commit_rollups(db, category):
 
     package.refresh_from_db()
 
-    assert package.last_commit_date == datetime.date(2022, 2, 20)
+    assert package.last_commit_date == datetime.datetime(2022, 2, 20, 2, 22)
 
     hist = _parse_histogram(package.commits_over_52)
     assert len(hist) == 52


package/models.py
```diff
@@ -169,6 +159,16 @@ class Package(BaseModel):
     last_exception_at = models.DateTimeField(blank=True, null=True)
     last_exception_count = models.IntegerField(default=0, blank=True, null=True)
 
+    commits_over_52 = models.TextField(_("Commit List Over 52 Weeks"), blank=True)
```
jefftriplett: Is there any advantage to this being a TextField vs. a JSONField that is defaulted to a list()?

saadmk11: I just renamed the commit_list field which was already a TextField, the PR is not ready yet, I was considering using ArrayField for this, but we can also do JSONField.

jefftriplett: Sorry, I'll let you keep working on it. 

If we are going to change it, I'd use a JSONFIeld + list() over an ArrayField or TextField.

saadmk11: Sounds good  👍

saadmk11:
closing in favor of https://github.com/djangopackages/djangopackages/pull/1513, this PR has many things unrelated to part 1

PR #1513, opened 2026-01-16
[Part 1] Move Frequently Used Stats to Package Model and Refactor PyPI Fetch Functionality

saadmk11:
**This PR (Part 1):**

1. Adds new fields to store frequently used stats on the Package Model (`commits_over_52w`, `last_commit_date`, `latest_version`)
2. Adds Django management command to sync the new fields from existing information. (ex: `uv run manage.py sync_package_stat_fields --batch-size 500`)
3. Adds new unified package updater management command (ex: `uv run manage.py package_updater2 --limit 300 --stale-hours 24 --jitter 0 --chunk-size 100 --git-max-per-minute 60 --pypi-max-per-minute 60 --time-budget 3300`
4. Added `scores.py` to calculate scores for packages with support for bulk update.
5. Updates `fetch_package_data_task` task to only save the package once to improve performance.
6. Added `package/stats_updater.py` to update the new stats fields (including `supports_python3`) with support of bulk operations for performance.
7. Added `pypi.py` file for fetching pypi packages. Adds better license handling as well.
8. Should not break any existing pages.

**Part 2 (Future):**
1. Remove unused fields from Package model.
2. Use the newly created Package Fields.
3. (Maybe) Enable the grid features to see how it performs.

### DIFF
--- a/package/tests/test_management_sync_stat_fields.py
+++ b/package/tests/test_management_sync_stat_fields.py
@@ -1,0 +1,131 @@
+import datetime
+
+import pytest
+from django.core.management import call_command
+from django.utils import timezone
+from model_bakery import baker
+
+from package.models import Commit, Package, Version
+
+
+@pytest.fixture
+def package_with_stats(db, category):
+    package = baker.make(
+        Package,
+        category=category,
+        title="Stats Package",
+        slug="stats-package",
+        repo_url="https://github.com/stats/package",
+        commit_count=0,
+        last_commit_date=None,
+        commits_over_52w=[],
+        latest_version=None,
+        supports_python3=False,
+    )
+
+    # Create some commits
+    now = timezone.now()
+    baker.make(Commit, package=package, commit_date=now - datetime.timedelta(days=1))
+    baker.make(Commit, package=package, commit_date=now - datetime.timedelta(days=10))
+    baker.make(
+        Commit, package=package, commit_date=now - datetime.timedelta(days=400)
+    )  # Older than 52 weeks
+
+    # Create versions
+    baker.make(
+        Version,
+        package=package,
+        number="1.0.0",
+        upload_time=now - datetime.timedelta(days=100),
+        hidden=False,
+        supports_python3=False,
+    )
+    v2 = baker.make(
+        Version,
+        package=package,
+        number="2.0.0",
+        upload_time=now - datetime.timedelta(days=10),
+        hidden=False,
+        supports_python3=True,
+    )
+    baker.make(
+        Version,
+        package=package,
+        number="3.0.0beta",
+        upload_time=now - datetime.timedelta(days=1),
+        hidden=True,
+    )  # Hidden
+
+    return package, v2
+
+
+def test_sync_package_stat_fields(package_with_stats):
+    package, latest_v = package_with_stats
+
+    call_command("sync_package_stat_fields")
+
+    package.refresh_from_db()
+
+    # Check Commit Stats
+    assert package.commit_count == 3
+    assert package.last_commit_date is not None
+    # The last commit date should be the most recent one (now - 1 day)
+    # We can check if it's close to now - 1 day, or just that it's updated.
+
+    # Check commits_over_52w
+    assert len(package.commits_over_52w) == 52
+    # We expect some commits in recent weeks.
+    # 1 day ago is week 0. 10 days ago is week 1.
+    assert package.commits_over_52w[51] >= 1  # week 0 (reversed)
+    assert package.commits_over_52w[50] >= 1  # week 1 (reversed)
+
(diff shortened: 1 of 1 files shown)


saadmk11:
@jefftriplett @ryancheley This PR is ready for your review.

saadmk11:
I am adding this [commit](https://github.com/djangopackages/djangopackages/pull/1513/commits/b4d35f6860ada165f0cdfdad351f5ae9176bc247) as a part of https://github.com/djangopackages/djangopackages/issues/1520 because this commit provides us with a more significant performance improvement and both of these are related.

saadmk11:
# Migration Plan: Package Statistics Optimization

This document outlines the plan to migrate, populate, and maintain the new optimized statistic fields on the `Package` model. These fields cache expensive calculations (commits, versions, python support) to improve read performance and ingestion performance.

## 1. Overview of Changes

We are transitioning from calculating the following values on-the-fly (or via expensive joins) to using cached fields on the `Package` model:

| Field | Type | Origin | Status |
|---|---|---|---|
| `commit_count` | `IntegerField` | Aggregated from `Commit` model | Added in Migration `0024` |
| `commits_over_52w` | `JSONField` | Computed from `Commit` model (last 52 weeks) | Added in Migration `0024` |
| `last_commit_date` | `DateTimeField` | Max `commit_date` from `Commit` model | Added in Migration `0024` |
| `latest_version` | `ForeignKey(Version)` | Latest visible `Version` (by upload time) | Added in Migration `0024` |
| `supports_python3` | `BooleanField` | From latest `Version` or classifiers | Existing (Migration `0008`) |

## 3. Migration & Backfill (One-time)

### Step 1: Apply Schema Migrations

The schema changes are captured in migration `0024_package_commit_count_package_commits_over_52w_and_more`.

Run the following command to apply the changes to the database:

```bash
python manage.py migrate package
```

### Step 2: Data Backfill (Population)

Once the columns exist, they will be empty. We need to populate them using the existing data in `Commit` and `Version` tables.

Use the `sync_package_stat_fields` management command. This command can process the entire database quickly using bulk updates.

**Recommended Command (Dry Run first):**

```bash
python manage.py sync_package_stat_fields --batch-size 1000 --dry-run
```

**Execute Backfill:**

```bash
python manage.py sync_package_stat_fields --batch-size 1000
```

*Note: The command uses `bulk_update` for efficiency. For a large number of packages, it may take some time but prints progress.*

### Step 3: Verification

1.  **Admin Panel:** Log in to the Django Admin and view a `Package`.

## 4. Ongoing Maintenance (Daily Updates)

To keep these fields up-to-date, we use `package_updater2` (which fetches new data and can update scores) and `calculate_score` (which purely recalculates scores).

### Command: `package_updater2`

This is the primary maintenance command. It fetches new data from PyPI and Git providers and automatically updates all statistic fields, **including scores**.

**Options:**
- `--stale-hours 24`: Only update packages that haven't been updated in 24 hours.
- `--time-budget 3600`: Run for max 1 hour then exit (useful for frequent cronjobs).
- `--chunk-size 200`: Batch size for DB writes.
- `--skip-scores`: Skips score recalculation if you want to do it separately (default is to update scores).

**Cron Strategy:**
With ~6,000 packages, we need to process approximately 250 packages per hour to ensure all packages are updated within a 24-hour window. Running hourly with a slightly higher limit (e.g., 300) ensures we hit this target while accounting for potential errors or skips.

```bash
# Run every hour, process packages older than 24h, limited to 300 per run.
# This ensures all 6,000 packages are updated every 24 hours.
0 * * * * python manage.py package_updater2 --stale-hours 24 --limit 300 --time-budget 3300 --chunk-size 100
```

### Command: `calculate_score` (Alternative / Supplemental)

This command purely recalculates the `score` field based on existing data in the database. It does **not** fetch external data.

**Use Case:**
- We want to refresh scores for *all* packages quickly without waiting for the fetcher to cycle through them.
- We used `--skip-scores` with `package_updater2`.
- We changed the scoring algorithm and need a global re-score.

**Options:**
- `--batch-size 1000`: Process 1000 packages at a time.

**Cron Strategy (Optional):**
If `package_updater2` covers scoring needs, this is optional.

```bash
# Optional: Run once a day to ensure all scores are fresh and consistent
0 4 * * * python manage.py calculate_score --batch-size 1000
```

## 5. Rollback Plan

If issues arise, we can revert the schema changes.

```bash
python manage.py migrate package 0023
```

## Next Steps (Part 2)

After successful migration and backfill, the codebase (views, models, templates) needs to be updated to read from these new fields instead of the old properties/methods. This will be covered in Part 2.

jefftriplett:
Btw, I'm still ok with merging this and working through anything that falls out. I don't suspect there will be much.

PR #1519, opened 2026-01-18
Bump gunicorn workers to 6 and django-q workers to 4

jefftriplett:
## Summary
- Increase gunicorn workers from 4 to 6
- Increase django-q workers from 2 to 4
- Better utilizes 4 CPU / 16GB machine resources


jefftriplett:
Server looks good, even with the dev server processes running. Bumping it up a bit to check CPU. 

I'm going to stop the dev process just to see how the CPU, etc works.

PR #1524, opened 2026-01-23
Fix link style in post detail and list templates

saadmk11:
Links were not visible:

Screenshot 2026-01-23 at 5 55 01 PM


saadmk11:
Fixed:
Screenshot 2026-01-23 at 6 09 58 PM

Issue #1520, opened 2026-01-19
[Proposal] Remove `Commit` Model due to Performance Bottleneck on Large Repositories


saadmk11:
## Summary
Currently, the platform fetches all commits from upstream repositories and creates corresponding commit objects in the database. This is causing severe performance bottlenecks, especially for repositories with thousands of commits (e.g., djangopackages/djangopackages has 4,926 commits). Thousands of database objects are created and stored, which is unnecessary after the upcoming changes at https://github.com/djangopackages/djangopackages/pull/1513; all we require are the last commit date, commit count, and the commit activity (52 weeks commit graph).

## Problem
- Fetching all commits via the paginated API and creating database objects for each commit leads to huge and unnecessary DB growth.
- Fetching all commits via the paginated API is really slow as for large projects we need to make a significant amount of API requests, causing rate-limits and timeouts.
- For large repositories, this causes slow updates and aggegations.
- With the new stats-driven feature set, the bulk commit objects are almost never used post-creation. (used to count the number of commits only)

## Proposal
- Stop creating individual commit DB objects per repository.
- Only fetch and store the following aggregate statistics without storing every commit:
    - Last commit date
    - Commit count (total)
    - Weekly commit graph (as an array of week totals for the last 52 weeks)
- All of these can be obtained via provider APIs (see code example).

## Analysis of Provided Code

- The included code efficiently minimizes API calls and database writes: 
    - **GitHub:** Uses the commits API with `per_page=1` to fetch the most recent commit and inspect the `Link` header for the commit count.  Fetches the weekly commit history from `/stats/participation` (pre-computed by GitHub).
    - **Aggregate queries mean:**
        - Only 2 API calls, regardless of total commit count
        - No need to loop over pages containing thousands of commits
        - Performance and bandwidth requirements are drastically reduced
    - Helper functions for GitLab, Bitbucket, and Codeberg perform similar fetches, but limitations and caveats exist.
      
## Caveats and Limitations for Other Providers

- **GitHub:**  
  - The code provided below is optimal; GitHub's participation stats endpoint gives accurate 52-week history.
  
- **GitLab:**  
  - No single API exposes the weekly commit histogram, so the code manually paginates all commits from the last year to build the graph (but not all commits).
  
- **Bitbucket:**
  - No total commit count endpoint.  The code must paginate through commits (could be expensive, same as current behavior).
  - The weekly histogram is similarly expensive since you need to fetch all commits in the last year and bin them manually. 
[...]

Issue #1498, opened 2026-01-13
Disabled commits sparkgraphs on grid detail page


jefftriplett:
## Summary

Temporarily disabled the commits sparkgraphs on the grid detail page to address performance issues.

## Changes Made

**Template (`templates/partials/grid_comparison_table.html`):**
- Commented out desktop commits row (lines 185-200)
- Commented out mobile commits section (lines 533-540)

**View (`grid/views.py`):**
- Commented out `Commit` import
- Commented out `cutoff` variable
- Commented out commits Prefetch query
- Commented out `commits_over_52` calculation and serialization

## To Re-enable

Uncomment the relevant sections in both files when ready to restore this functionality.

jefftriplett:
cc @saadmk11 

Trying this out before we build out the recalculation logic.

saadmk11:
Enabled in production

Issue #1497, opened 2026-01-13
Remove Python3 Filter from GridDetail Page


saadmk11:
I don't mind if we lose the python3 filter all together. It's outlined its usefulness by ~12 years. 

_Originally posted by @jefftriplett in https://github.com/djangopackages/djangopackages/pull/1496#discussion_r2686575833_

jefftriplett:
I removed a few of these while I was turning off extra queries.

saadmk11:
We can close this issue as all the related code has been removed

PR #1532, opened 2026-02-01
[Part 2] Use new Package model fields to improve performance

saadmk11:
**Use new Package model fields to improve performance**

- Removed unused Commit model references in views and builders.
- Updated package detail views to utilize latest_version instead of last_released.
- Optimized queries in PackageRulesView and PackageDetailView by adding select_related for latest_version.
- Adjusted package weight calculation to use latest_version and last_commit_date.
- Updated templates to reflect changes in package attributes, including latest_version_number and last_commit_date.
- Enhanced grid comparison and package listing templates to use package directly.

**Next Steps**

- Remove unused Commit Model
- https://github.com/djangopackages/djangopackages/issues/1500 was not updated. We should first observe the state after this is deployed then we can add back features if the results are satisfactory.

**Related**
- https://github.com/djangopackages/djangopackages/issues/1520
- https://github.com/djangopackages/djangopackages/pull/1513
- https://github.com/djangopackages/djangopackages/issues/1497
- https://github.com/djangopackages/djangopackages/issues/1498

### DIFF
--- a/templates/partials/site_notice.html
+++ b/templates/partials/site_notice.html
@@ -7,7 +7,7 @@
             <i class="flex-shrink-0 ph-bold ph-warning"></i>
             <span>
                 {% blocktranslate %}
-                    We've temporarily disabled some Grid and Package features while we work on stability improvements. Thanks for your patience!
+                    We've temporarily disabled some Grid features while we work on stability improvements. Thanks for your patience!
                 {% endblocktranslate %}
             </span>
         </div>


jefftriplett:
@saadmk11 I'm comfortable trusting you and your gut for when this is ready. I'll defer a pass by @ryancheley but with our new setup, we can patch/fix any small issues. Well don on this as always.

saadmk11:
> @saadmk11 I'm comfortable trusting you and your gut for when this is ready. I'll defer a pass by @ryancheley but with our new setup, we can patch/fix any small issues. Well don on this as always.

Thanks, made some small adjustments, but it is ready from my end. :)

PR #1540, opened 2026-02-14
Migrate to Prek and Use GitHub Actions for Auto-Update

saadmk11:


templates/stack.html
```diff
@@ -115,10 +115,6 @@ <h2 class="mb-4 text-2xl font-bold text-foreground">{% translate "Third-party Se
                                     <td class="py-3 pr-4 font-medium text-foreground">Read the Docs</td>
                                     <td class="py-3 text-muted-foreground">Documentation Hosting</td>
                                 </tr>
-                                <tr class="transition-colors hover:bg-muted/30">
-                                    <td class="py-3 pr-4 font-medium text-foreground">Pre-Commit CI</td>
-                                    <td class="py-3 text-muted-foreground">Continuous Integration</td>
-                                </tr>
```
saadmk11: If we migrate we need to remove pre-commit.ci integration (c.c: @jefftriplett)

saadmk11:
This PR does not implement auto-fixing on PR's as doing it for forks is quite difficult and not recommended. 

This PR was experiment to see how prek works and we can keep using our existing workflow as it is working quite well for us.

pauloxnet:
Well done 👍

PR #1591, opened 2026-04-22
Support multiple simultaneous site banners

jefftriplett:
Getting ready to exercise this a bit. 

## Summary

- `get_active_banner()` → `get_active_banners()` — now returns a list of all active banners instead of just the first one
- Context processor passes `active_banners` (list) to templates; dismissed banners are filtered out individually from the list
- `banner.html` loops over `active_banners`; each banner gets a unique `id="site-banner-{{ banner.pk }}"` so HTMX dismiss targets the correct one
- Updated tests to cover multiple simultaneous banners and partial dismissal
- Style anchor tags in banner content with underline

### DIFF
--- a/banners/cache.py
+++ b/banners/cache.py
@@ -2,27 +2,27 @@ from django.conf import settings
 from django.core.cache import cache
 from banners.models import Banner
 
-BANNER_CACHE_KEY = "banners:active_banner"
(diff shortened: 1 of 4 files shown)


saadmk11 (approved):
Looks good to me 👍

PR #1594, opened 2026-04-24
Fix RTD build: switch to uv run via build.commands

jefftriplett:
## Summary

- RTD's native uv support (`method: uv, command: sync`) installs dependencies correctly but then invokes `python -m mkdocs build` using `/bin/python`, which doesn't exist on Ubuntu 24.04
- Switches to `build.commands` with `uv run mkdocs build` so uv handles venv activation — no `/bin/python` lookup involved
- Also bumps the Python tool version to `3.14` to match the RTD blog post example

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,7 +10,15 @@ formats: all
 build:
   os: ubuntu-24.04
   tools:
-    python: "3.14"
-  commands:
-    - uv sync --group docs
-    - uv run mkdocs build --clean --site-dir $READTHEDOCS_OUTPUT/html --config-file docs/mkdocs.yml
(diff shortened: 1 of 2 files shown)


saadmk11:
I think this is currently only supported for sphinx builds: 

https://github.com/readthedocs/readthedocs.org/blob/b8e2239c7383fc4874c446091f2c962b986c1228/readthedocs/doc_builder/backends/sphinx.py#L179-L191

saadmk11:
We can just revert the merged PR for now

jefftriplett:
Related issue: https://github.com/djangopackages/djangopackages/issues/1595

PR #1626, opened 2026-07-04
Fix `/health/` -> `500` by configuring the Redis check for `django-health-check` `4.x`

saadmk11:
While investigating a [Sentry issue](https://djangopackages.sentry.io/issues/7376044436/events/d4bd47db3fd6448cb7a5b9feabefe3c3/) I found that every request to `/health/` has been returning a 500 since [#1581](https://github.com/djangopackages/djangopackages/pull/1581) bumped django-health-check from 3.20 to 4.x. The new version's Redis check no longer reads `REDIS_URL` from settings and raises a `ValueError` at instantiation unless it's given a `client_factory`, and the upgrade left it registered without one, crashing the whole endpoint on every request. I've configured the check the way the 4.x docs recommend.

**Ref:** 

1. https://django-packages.sentry.io/issues/7376044436/events/d4bd47db3fd6448cb7a5b9feabefe3c3/
2. https://github.com/codingjoe/django-health-check/issues/632

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -41,3 +41,5 @@ CACHES = {
         "BACKEND": "django.core.cache.backends.dummy.DummyCache",
     }
 }
+
+REDIS_URL = "redis://localhost:6379/0"
--- a/tests/test_health.py
+++ b/tests/test_health.py
@@ -1,0 +1,20 @@
+from unittest.mock import AsyncMock, patch
+
+from django.test import override_settings
+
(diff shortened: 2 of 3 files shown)


ryancheley (approved):
Looks good to me. I had a similar issue on a personal project and the changes here track with what I had to do. 

Great job @saadmk11 💯

jefftriplett (approved):

PR #1628, opened 2026-07-07
Fix GitHub Login Issue (`social-app-django` 6.0 breaking change)

saadmk11:
`social-auth-app-django` **6.0** added a breaking change that made the login views POST-only, so every GET path into `/auth/login/github/` (navbar links, `login_required` redirects etc.) now returns HTTP 405.

- Replaced the navbar "Log In" links (desktop and mobile) with forms that POST to `social:begin` with a CSRF token.
- Added a static `/login/` page in `urls.py` that renders the same form. Redirects can only GET, so `login_required` needs a page that carries the POST form. It passes `?next=` through as a hidden input.
- Pointed `LOGIN_URL` at `/login/`.

**Reference:**

> ### Breaking
> - Login views now always require POST requests, and the `SOCIAL_AUTH_REQUIRE_POST` setting was removed.

https://github.com/python-social-auth/social-app-django/blob/master/CHANGELOG.md#breaking

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -227,7 +227,7 @@ ABSOLUTE_URL_OVERRIDES = {
 
 AUTH_PROFILE_MODULE = "profiles.Profile"
 
-LOGIN_URL = "/auth/login/github/"
+LOGIN_URL = "/login/"
 LOGIN_REDIRECT_URLNAME = "home"
 LOGOUT_REDIRECT_URL = "/"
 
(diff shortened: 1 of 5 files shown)


saadmk11:
We can ignore prek for now as version 0.25.0 of rustywind introduces some changes that reformats almost all of the HTML files (mostly re-ordering tailwind classes), we can add another PR that can do that.

https://github.com/avencera/rustywind/blob/master/CHANGELOG.md#0250-alpha1---2026-05-30

jefftriplett (approved):
Nice catch

PR #1629, opened 2026-07-07
Reformat HTML templates with rustywind 0.25.0

jefftriplett:
Follow-up to #1628.

rustywind 0.25.0 ([changelog](https://github.com/avencera/rustywind/blob/master/CHANGELOG.md#0250-alpha1---2026-05-30)) re-orders Tailwind classes, which reformats almost all of the HTML templates. As discussed in [this comment](https://github.com/djangopackages/djangopackages/pull/1628#issuecomment-4903020553), that reformat was split out of #1628 into its own PR to keep the diff reviewable.

Changes are purely Tailwind class re-ordering (no behavior change), produced by running the `rustywind` pre-commit hook across `templates/`.

### DIFF
--- a/.github/workflows/prek-autoupdate.yml
+++ b/.github/workflows/prek-autoupdate.yml
@@ -20,6 +20,7 @@ jobs:
         uses: j178/prek-action@v2
         with:
           install-only: true
+          prek-version: 0.4.8
 
       - name: Run prek auto-update
         run: prek auto-update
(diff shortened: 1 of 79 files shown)


jefftriplett:
8 templates were restored and excluded from the rustywind hook here because rustywind mangles Django template tags (`{% if %}`/`{% elif %}`/`{% cycle %}`) embedded inside `class="..."` attributes — this reproduces on every rustywind version, not just 0.25.0. Tracking the long-term refactor (lift the conditionals out of the class attributes so they can be sorted, then drop them from the exclude list) in #1630.

Issue #1456, opened 2025-11-14
:tractor: Migrate from requests to httpx


jefftriplett:
The most annoying part of this will be the mocked requests parts of our code. 

I would start with these two to start: 
- [ ] `core/utils.py`
- [ ] `package/models.py`

Then I'd dive a little deeper into the other areas of code that are using it.

Divan009:
I would like to try this out @jefftriplett

jefftriplett:
@Divan009 you are welcome to work on any issue that's open and not assigned. PRs and issue comments are always welcomed here.

jefftriplett:
Done in #1662, which moved the codebase from requests to httpx2. No code imports `requests` any more, and it's no longer a dependency in `pyproject.toml`.

PR #1662, opened 2026-09-14
Migrate from requests to httpx2

mounirmesselmeni:
Refs #1456 

The main requests usage was migrated to httpx in the actual runtime code

I also updated the tests and mock setup to match the new HTTP client, including the pytest_httpx fixture usage.


pyproject.toml
```diff
@@ -94,6 +94,7 @@ dependencies = [
     "Django>5.0",
     "djangorestframework<3.19",
     "emoji",
+    "httpx",
```
pauloxnet: I would suggest to jump directly to httpx2 instead o migrating to httpx and then migrato also to httpx2
https://github.com/pydantic/httpx2

mounirmesselmeni: I agree, will update the PR

pyproject.toml
```diff
@@ -139,6 +138,7 @@ dev = [
     "pytest",
     "pytest-cov",
     "pytest-django",
+    "pytest-httpx2",
```
pauloxnet: This packages seems to have 3 commit only. Have you considered to check this one https://github.com/angryfoxx/httpx2-pytest ?
Disclaimer: I haven't used any of them.

jefftriplett:
@mounirmesselmeni Thank you for the PR and this. I've been thinking about making this change for a while, but I keep getting held up because so many libraries require `httpx` or `requests`, and they both get installed either way. 

One example is https://github.com/mounirmesselmeni/djangopackages/blob/43b017bc48110c12d0bb820f5d9cf5d1606bab06/uv.lock#L264, which still requires `httpx` either way. 

Maybe having all three isn't that big of a deal, but it kind of annoys me that we don't have a better packing workaround for using x instead of y or z when these libraries ship via third-party packages. 

Let me stew on it for a day, but I think we'll move forward with it.

jefftriplett:
@mounirmesselmeni awesome work. thank you for doing the work!

@pauloxnet thank you for the suggestion and assist!

PR #1672, opened 2026-09-17
Bump rustywind to 0.28.0 and drop the template exclude list

### LINKED ISSUE #1630: rustywind can't sort Tailwind classes containing Django template tags (8 templates excluded)
I had Claude Code note this for future follow-up. 

## Background

rustywind (the Tailwind class sorter run via the `rustywind` pre-commit hook) cannot safely sort Tailwind classes that contain Django template tags inside a `class="..."` attribute — e.g. `class="border-b {% if x %}bg-violet-50{% elif y %}bg-sky-50{% endif %}"` or `{% cycle %}` used for classes.

When it encounters these, it treats the `{% if %}` / `{% elif %}` / `{% cycle %}` content as Tailwind tokens and reorders them, stripping the `{%`/`%}` delimiters. The result is either a `TemplateSyntaxError` (e.g. `Invalid block tag 'endfor', expected 'endif'`) or — worse — a template that still compiles but renders literal garbage (`elif 'maybe' %}`) and silently drops the conditional so every class is always applied.

This is not version-specific — it reproduces on rustywind 0.21.0, 0.24.0, and 0.25.0.

## Current state (workaround in #1629)

In #1629 these templates were **restored to their original markup and excluded** from the rustywind hook via `exclude:` in `.pre-commit-config.yaml`, so pre-commit / pre-commit.ci / prek all skip them. This keeps them valid but means their Tailwind classes are not auto-sorted.

Affected templates:

- `templates/homepage/index.html` (`{% cycle %}`)
- `templates/homepage/readiness_detail.html`
- `templates/partials/_ethicalads-tag.html`
- `templates/partials/banner.html`
- `templates/partials/form_renderer.html`
- `templates/partials/grid_comparison_table.html`
- `templates/partials/messages.html`
- `templates/partials/suggestions.html`

## Proposed long-term fix

Refactor the conditional logic out of the `class="..."` attributes so rustywind can sort the static classes, then drop each file from the `exclude:` list. Options per case:

- Move the `{% if %}`-selected classes into a template variable / `{% with %}` computed above the element, then interpolate a single `{{ classes }}`.
- Or push the branch-specific classes into small partials / a template filter / a helper that returns the class string.

This changes rendering structure, so it should be done deliberately with visual verification, not folded into a formatting PR.

## Acceptance

- [ ] Each of the 8 templates renders identically to today.
- [ ] Each is removed from the rustywind `exclude:` list.
- [ ] `rustywind` hook passes on all templates.

jefftriplett:
## What

rustywind **0.26.0** fixed the bug behind #1630: sorting Tailwind classes no longer mangles Django template tags inside `class="..."`. The eight templates excluded in #1629 can come back under the hook.

- `rustywind@0.25.0` → `rustywind@0.28.0`
- Removed the `exclude:` block from the `rustywind` hook
- Reformatted the 6 previously-skipped templates that needed sorting

Note that rustywind is a `local` hook using `additional_dependencies`, so `prek autoupdate` never bumps it — it had drifted three minors behind.

## Verification

- Reproduced the original corruption on 0.25.0 and confirmed 0.26.0/0.27.0/0.28.0 all leave the template tags intact.
- Compared the full `{% ... %}` / `{{ ... }}` token sequence before and after for every changed template — byte-identical, so only class ordering moved.
- Loaded every template through Django's template engine: 0 syntax errors.
- `prek run rustywind --all-files` passes.
- `just pytest`: 369 passed, 1 skipped, 3 xfailed.

Only the 6 formerly-excluded files changed; no other template moved, so the 0.25→0.28 upgrade did not shift sort order for already-sorted files.

Closes #1630


saadmk11 (approved):
LGTM! 💯

PR #1675, opened 2026-09-18
Use repo homepage as a fallback for documentation_url

### LINKED ISSUE #1666: Use repo homepage as a fallback for documentation_url
> [!NOTE]
> This issue was LLM generated.

## Problem

When a package's PyPI metadata has no documentation link, `documentation_url` stays empty even when the repository lists a website.

Example: [django-probe](https://github.com/tim-schilling/django-probe)

- PyPI `project_urls` only has `Repository`, and `docs_url` / `home_page` are empty.
- The GitHub repo's `homepage` is set to https://djangoprobe.org.

Right now `GitHubHandler.fetch_metadata` never reads `homepage`, so we miss the link.

## Proposal

Fill `documentation_url` from the first match below:

1. PyPI `docs_url`, or a `Documentation`/`Docs` entry in `project_urls` (what we check today)
2. PyPI `Homepage` entry in `project_urls`
3. The repo's website field (GitHub `homepage`, Forgejo/Codeberg `website`)

## Notes

- Only fill `documentation_url` when it's blank, so we don't overwrite a URL someone entered by hand. `package/pypi.py` currently overwrites it on every fetch whenever PyPI has a docs link.
- Skip homepage values that point back to the repo or to pypi.org.
- Match `project_urls` keys case-insensitively.

cmanjah:
Closes #1666

When a package has no documentation URL, fill it from:

1. PyPI docs / Documentation links
2. The PyPI Homepage project URL
3. The GitHub homepage or Forgejo/Codeberg website field

Homepages that point back at the repo or at PyPI are skipped. The field is only filled when it is blank, so a URL entered by hand is left alone.


package/pypi.py
```diff
@@ -307,8 +319,11 @@ def update_package_from_pypi(
     if pypi_info.supports_python3 is not None:
         package.supports_python3 = pypi_info.supports_python3
 
-    if pypi_info.docs_url:
-        package.documentation_url = pypi_info.docs_url
+    if not package.documentation_url:
+        if pypi_info.docs_url:
+            package.documentation_url = pypi_info.docs_url
+        else:
+            maybe_set_documentation_url(package, pypi_info.homepage_url)
```
jefftriplett, line 326: Can we keep PyPI authoritative here? The blank check covers `docs_url` too, not just the new fallback, so `documentation_url` only ever gets written once. If a project moves its docs we keep serving the old URL forever.

`maybe_set_documentation_url` already bails when the field is set, so the fallback stays blank-only on its own.

```suggestion
    if pypi_info.docs_url:
        package.documentation_url = pypi_info.docs_url
    else:
        maybe_set_documentation_url(package, pypi_info.homepage_url)
```

`test_does_not_overwrite_existing_documentation_url` will need updating to match.

jefftriplett:
Looks great. Thanks for the contribution! I added one note to think about.

package/pypi.py
cmanjah: You are right! Done. PyPI `docs_url` is authoritative again. The homepage fallback still only fills a blank field

